PR #174, opened 2014-05-22
Use X-Forwared-For when behind reverse-proxy

mattikus:
Allow the use of [X-Forwarded-For](http://en.wikipedia.org/wiki/X-Forwarded-For) if the header exists to allow reverse-proxies and loadbalancers to be set up in front of a bcfg2-server.

So far this diff only applies to the Builtin core. I could investigate applying it to the others as well, if there's interest in this PR.

### DIFF
--- a/src/lib/Bcfg2/Client/Proxy.py
+++ b/src/lib/Bcfg2/Client/Proxy.py
@@ -242,6 +242,7 @@ class XMLRPCTransport(xmlrpclib.Transport):
         self.scns = scns
         self.timeout = timeout
         self.protocol = protocol
+        self.cookie = None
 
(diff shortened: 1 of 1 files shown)


AlexanderS:
Anything new on this topic? As far as I can see the TODO items from above are still open?

mattikus:
Apologies, I no longer work somewhere where Bcfg2 is used and have since not had a need to continue work on this PR.

It was mainly meant for scaling the environment at my previous job. It probably needs refactoring and rebasing against the latest stable and some documentation on the parameters you'd pass.

Feel free to close this PR as well, as I do not believe I'll have the time or interest to finish it.

PR #244, opened 2014-11-25
Bundler: add modification support to Bundle dependencies

AlexanderS:
Bundle dependencies are now realized with RequiredBundle and support inheritance of the modification flag. This requires new client support and will only work with clients >= 1.4.0pre2.

The new tag name is required for validation. The tag name and type association in the schema has to be unambiguous and we need nested bundle tags for xincludes. Bundle dependency are only available in 1.4.0pre1, so I do not know if we need a migration script that renames the empty bundle tags to RequiredBundle.


solj:
Merged, sorry about the delay.

PR #250, opened 2014-12-09
Remove blanket excepts from plugins and lint

### LINKED ISSUE #249: Fix blanket exceptions
This is partly in response to #246 . We need to go through the code and remove as many of these as is reasonable.

stpierre:
This removes most blanket except: clauses from all plugins, including
the base plugin libraries, and bcfg2-lint. The few that remain should
all be necessary.

Most of the changes were quite minor, but this did require some
restructuring of the CfgPrivateKeyCreator; as a result, the tests for
that module were rewritten.

This partially fixes #249.

### DIFF
--- a/src/lib/Bcfg2/Server/Lint/GroupPatterns.py
+++ b/src/lib/Bcfg2/Server/Lint/GroupPatterns.py
@@ -4,7 +4,8 @@
 import sys
 
 from Bcfg2.Server.Lint import ServerPlugin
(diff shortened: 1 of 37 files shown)


stpierre:
@solj @AlexanderS @fennm Any comments on this before I merge it?

fennm:
This is a big change, but it looks like just the cleanup that is good for master.

solj:
LGTM :+1:

PR #252, opened 2014-12-13
Server/Core: semi-automatically handle list args with rmi

AlexanderS:
Our current implementation for "bcfg2-admin xcmd" does not support list
arguments for rmi calls. With the new decorater in Server.Plugin.helpers
all arguments are delivered as list to the rmi function.


AlexanderS:
I think it is better to implement the list handling in xcmd. I will create a new pull request for this.

PR #254, opened 2015-01-14
Plugins/Packages: fix initialization of YumSource

AlexanderS:
During ``__init__`` of the parent class ``get_repo_name`` is called. That needs
fields (``pump_id``) that are defined later in the ``__init__`` of YumSource.
We introduce the new function ``_init_attributes`` to parse all attributes
out of the ``<Source>`` tag before calling any other functions.

### DIFF
--- a/src/lib/Bcfg2/Server/Plugins/Packages/Source.py
+++ b/src/lib/Bcfg2/Server/Plugins/Packages/Source.py
@@ -143,6 +143,83 @@ class Source(Bcfg2.Server.Plugin.Debuggable):  # pylint: disable=R0902
         #: source
         self.essentialpkgs = set()
 
+        self._init_attributes(basepath, xsource, setup)
+
+        #: A set of all package names in this source.  This will not
+        #: necessarily be populated, particularly by backends that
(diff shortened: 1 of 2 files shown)


solj:
@davisj can you see if this fixes your issue?

stpierre:
@davisj Have you had a chance to test this out yet? Thanks!

davisj:
will try to get it tested today. Sorry for the wait

davisj:
Confirmed. This patch solved my problem. Thanks!

stpierre:
Could this be fixed more simply by calling the superclass ``Source.__init__()`` after the ``YumSource.__init__()`` sets ``pulp_id`` and other stuff?

AlexanderS:
@stpierre It is not that trivial... The pulp code in `YumSource.__init__` may override `self.arches`. Therefore the initialization code from the base class has to run before that. But in `Soruce.__init__` `self.arches` is already used to create the `url_map`. So the pulp stuff has to be executed in between.

stpierre:
Thanks for clarifying. LGTM then.

Issue #253, opened 2014-12-17
bcfg2-server -d does not drop root privileges


madduck:
If I specify `user=bcfg2` in the config, then this causes the server process to drop privileges… unless I run it with `-d`, in which case it continues to run as root.

PR #256, opened 2015-01-16
Server/Core: drop privileges even if not running as daemon

AlexanderS:
As requested in #253 this let the server to drop privileges even if not running as daemon (but if running as root).

### DIFF
--- a/src/sbin/bcfg2-info
+++ b/src/sbin/bcfg2-info
@@ -726,6 +726,8 @@ Bcfg2 client itself.""")
 
     def run(self, args):  # pylint: disable=W0221
         try:
+            if os.getuid() == 0:
+                self._drop_privileges()
             self.load_plugins()
             self.block_for_fam_events(handle_events=True)
             if args:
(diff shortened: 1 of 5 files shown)


solj:
@madduck Can you give this a shot when you get a chance? If it works as expected, :+1: from me.

src/sbin/bcfg2-info
```diff
@@ -738,6 +740,11 @@ Bcfg2 client itself.""")
     def _daemonize(self):
         pass
 
+    def _drop_privileges(self):
+        os.setgid(self.setup['daemon_gid'])
+        os.setuid(self.setup['daemon_uid'])
+        print("Dropped privileges to %s:%s" % (os.getuid(), os.getgid()))
```
stpierre: I don't like this ``print()`` -- it could well interfere with the output of ``bcfg2-info``.

stpierre:
There are three different ways to drop privs in this patch:

1. In ``BuiltinCore``, using ``daemon``;
2. In ``CherryPyCore``, using CherryPy;
3. In ``bcfg2-info``, using ``os.set[ug]id``.

Would it be possible and/or reasonable to make the ``daemon`` implementation the default (i.e., in ``BaseCore`` so that both ``BuiltinCore`` and ``bcfg2-info`` could use the same privilege dropping implementation? I'm not even sure if that'd work, but it'd be nice to limit the differences between ``bcfg2-info`` and the builtin core.

AlexanderS:
The daemon variant is not the default. This also removes the print in bcfg2-info because it now always uses logger.debug in BaseCore.

stpierre:
pylint fixes have been backported to maint, so you should be able to rebase this.

AlexanderS:
Rebased to fix build.

PR #257, opened 2015-01-27
Make SSHbase also a Connector plugin.

AlexanderS:
Make the public host key data avaialabe for the templates. I do not know if there is something special to do for metadata caching.


stpierre:
LGTM.

solj:
Merged, thanks!

PR #258, opened 2015-01-27
DBSettings: Fix time_zone option.

AlexanderS:
In the documentation, in the 1.3.x version and in django this option
is called time_zone (and not timezone). There is no reason to change this.

### DIFF
--- a/src/lib/Bcfg2/DBSettings.py
+++ b/src/lib/Bcfg2/DBSettings.py
@@ -109,7 +109,7 @@ def finalize_django_config(opts=None, silent=False):
             OPTIONS=opts.reporting_db_opts,
             SCHEMA=opts.reporting_db_schema)
 
-    settings['TIME_ZONE'] = opts.timezone
+    settings['TIME_ZONE'] = opts.time_zone
 
     settings['TEMPLATE_DEBUG'] = settings['DEBUG'] = \
         opts.web_debug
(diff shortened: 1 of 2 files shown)


src/lib/Bcfg2/DBSettings.py
```diff
@@ -261,7 +261,7 @@ class _OptionContainer(object):
             type=Bcfg2.Options.Types.comma_dict, default=dict()),
         # Django options
         Bcfg2.Options.Option(
-            cf=('reporting', 'timezone'), help='Django timezone'),
+            cf=('reporting', 'time_zone'), help='Django timezone'),
```
stpierre: Can we add a ``deprecated_cf`` here so that ``reporting.timezone`` will still work?

AlexanderS:
Something like that?

stpierre:
Urgh, it looks like {{deprecated_cf}} must have gotten removed in 1.4. And it's not the easiest thing to add back, either.

@solj, any suggestions for this?

AlexanderS:
Oh, then I think I will remove the deprecated patch. The ``timezone`` option was only in 1.4.0pre1, so it should not be critical to support this option.

stpierre:
Okay, sounds good to me. +1

solj:
Looks good to me as well. Merged. Thanks!

PR #259, opened 2015-01-27
Fix date format

AlexanderS:
I wanted to customize the date and time formats used in the reporting web interface. This is the way to do it with django. I could now use something like ``django_settings = USE_L10N=true, LANGUAGE_CODE=de-DE`` and get correct German date formats.

### DIFF
--- a/doc/man/bcfg2.conf.txt
+++ b/doc/man/bcfg2.conf.txt
@@ -718,6 +718,11 @@ Reporting options
         Maximum number of children for the reporting collector. Use 0 to
         disable the limit. (default is 0)
 
+    django_settings
+        Arbitrary options for the Django installation. The value expected
+        is a literal python dictionary, that is merged with the already set
+        django settings.
+
 See Also
 --------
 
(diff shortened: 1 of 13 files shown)


AlexanderS:
``Connection reset by peer`` in this one specific travis build should not be caused by this patch.

stpierre:
This looks good (and the test failure indeed seems unrelated). Could you add an example of using the new option to the documentation, perhaps in ``doc/reports/dynamic.txt``? Thanks!

solj:
Yep, restarted the build and it completed successfully. This looks good to me :+1: . Can you also mention this change in the 1.4.0pre2 release notes?

AlexanderS:
If we merge 257eb0c17 into master, the format of this option will change. So lets keep this open until we merge again. I will then update this pull request.

AlexanderS:
This is now on top of #295.

solj:
Merged, thanks!

AlexanderS:
Oh, just noticed, that I was missing the release notes again :(

PR #260, opened 2015-02-16
Plugin homebrew

jcftang:
This a first pass at teaching bcfg2 about homebrew on OSX. Much of this is lifted from the MacPorts.py file.

### DIFF
--- a/src/lib/Bcfg2/Client/Tools/HomeBrew.py
+++ b/src/lib/Bcfg2/Client/Tools/HomeBrew.py
@@ -1,0 +1,54 @@
+"""This provides Bcfg2 support for homebrew packages."""
+
+import Bcfg2.Client.Tools
+
+
+class HomeBrew(Bcfg2.Client.Tools.PkgTool):
+    """Homebrew package support."""
+    name = 'HomeBrew'
+    __execs__ = ["/usr/local/bin/brew"]
+    __handles__ = [('Package', 'homebrew')]
+    __req__ = {'Package': ['name', 'version']}
+    pkgtype = 'homebrew'
+    pkgtool = ('/usr/local/bin/brew install %s', ('%s', ['name']))
+
+    def RefreshPackages(self):
+        """Refresh memory hashes of packages."""
+        pkgcache = self.cmd.run(["/usr/local/bin/brew",
+                                 "list", "--versions"]).stdout.splitlines()
+        self.installed = {}
+        for pkg in pkgcache:
+            pkgname, version = pkg.strip().split()
+            self.logger.debug(" pkgname: %s version: %s" % (pkgname, version))
+            self.installed[pkgname] = version
+
+    def VerifyPackage(self, entry, _):
+        """Verify Package status for entry."""
+        self.logger.debug("VerifyPackage: %s : %s" % (entry.get('name'),
+                                                      entry.get('version')))
+
+        if entry.attrib['name'] in self.installed:
+            if (self.installed[entry.attrib['name']] ==
+                entry.attrib['version'] or
+                entry.attrib['version'] == 'any'):
+                return True
+            else:
+                self.logger.info("  %s: Wrong version installed.  "
+                                 "Want %s, but have %s" %
+                                 (entry.get("name"),
+                                  entry.get("version"),
+                                  self.installed[entry.get("name")],
+                                  ))
+
+                entry.set('current_version', self.installed[entry.get('name')])
+                return False
+        entry.set('current_exists', 'false')
+        return False
+
+    def Remove(self, packages):
+        """Remove extra packages."""
+        pkg_names = [p.get('name') for p in packages]
+        self.logger.info("Removing packages: %s" % pkg_names)
+        self.cmd.run(["/usr/local/bin/brew", "uninstall"] + pkg_names)
+        self.RefreshPackages()
+        self.extra = self.FindExtra()
--- a/doc/releases/1.4.0pre2.txt
+++ b/doc/releases/1.4.0pre2.txt
@@ -18,6 +18,7 @@ That said, please help us test the release in non- and preproduction
 environments.
 
 * NagiosGen: Add bundles to configuration
+* HomeBrew: Initial add of plugin
 
 backwards-incompatible user-facing changes
 ------------------------------------------


AlexanderS:
You may want to drop the first 6 commits from this pull request, because they are already in master.

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def RefreshPackages(self):
+        """Refresh memory hashes of packages."""
+        pkgcache = self.cmd.run(["/usr/local/bin/brew",
+                                 "list", "--versions"]).stdout.splitlines()
+        self.installed = {}
+        for pkg in pkgcache:
+            pkgname = pkg.split(' ')[0].strip()
+            version = pkg.split(' ')[1].strip()
+            self.logger.info(" pkgname: %s version: %s" % (pkgname, version))
```
stpierre: This looks like it should be logged at a debug level, if at all.

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,82 @@
...
+        self.RefreshPackages()
+
+    def RefreshPackages(self):
+        """Refresh memory hashes of packages."""
+        pkgcache = self.cmd.run(["/usr/local/bin/brew",
+                                 "list", "--versions"]).stdout.splitlines()
+        self.installed = {}
+        for pkg in pkgcache:
+            pkgname = pkg.split(' ')[0].strip()
+            version = pkg.split(' ')[1].strip()
```
stpierre: ``pkgname, version = pkg.strip().split()``

Or, if version might contain spaces:

``pkgname, version = pkg.strip().split(None, 2)``

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,82 @@
...
+            self.logger.info(" pkgname: %s version: %s" % (pkgname, version))
+            self.installed[pkgname] = version
+
+    def VerifyPackage(self, entry, _):
+        """Verify Package status for entry."""
+	self.logger.info("VerifyPackage: %s : %s" % (entry.get('name'),
+                                                     entry.get('version')))
```
stpierre: Log this at debug.

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,82 @@
...
+        if 'version' not in entry.attrib:
```
stpierre: This block of code is not reachable; ``Verify*`` is only called if ``canVerify`` succeeds, and ``canVerify`` checks that all attributes in ``self.__req__`` are present.

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,82 @@
...
+                entry.set('current_version', self.installed[entry.get('name')])
+                return False
+        entry.set('current_exists', 'false')
+        return False
+
+    def Remove(self, packages):
+        """Remove extra packages."""
+	pkgline = ""
+	for pkg in packages:
+		pkgline += " " + pkg.get('name')
```
stpierre: ``pkgline = " ".join(p.get('name') for p in packages)``

stpierre: Actually, see below.

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,82 @@
...
+
+        self.logger.info("Removing packages: %s" % pkgline)
+        self.cmd.run("/usr/local/bin/brew uninstall %s" % pkgline)
```
stpierre: Use the list version of ``run()`` instead:

    pkg_names = [p.get('name') for p in packages]
    self.logger.info("Removing packages: %s" % pkg_names)
    self.cmd.run(["/usr/local/bin/brew", "uninstall"] + pkg_names)

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,82 @@
...
+	"""Install packages."""
+	pkgline = ""
+	for pkg in packages:
+		pkgline += " " + pkg.get('name')
+
+	self.logger.info("About to install packages : " + pkgline)
+
+	try:
+		self.logger.info("Running : brew install %s" % pkgline)
+		self.cmd.run("/usr/local/bin/brew install %s" % pkgline)
```
stpierre: Use a list comprehension to build your list of package names, and the list version of ``run()`` here as well.

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,82 @@
...
+        self.RefreshPackages()
+        self.extra = self.FindExtra()
+
+    def InstallPackage(self, packages, states):
+	"""Install packages."""
+	pkgline = ""
+	for pkg in packages:
+		pkgline += " " + pkg.get('name')
+
+	self.logger.info("About to install packages : " + pkgline)
```
stpierre: You've already got a much better log message on line 78 below; this can be removed.

stpierre:
Aside from some fairly minor code issues, LGTM.

jcftang:
The comments sound reasonable, I'll fix up the patch and resubmit it.

src/lib/Bcfg2/Client/Tools/HomeBrew.py
jcftang: I'm not entirely sure what you suggest to do to resolve this particular issue. Shall I just remove this method?

jcftang:
I think I've addressed most of the comments from the above. However after rebasing to the master branch...

```
(bcfg2)jtang@awesome-o:bcfg2$ bcfg2 -C etc/bcfg2.conf  -v -q  -c tmp
Starting Bcfg2 client run at 1424109059.87
Failed to instantiate tool <class 'Bcfg2.Client.Tools.HomeBrew.HomeBrew'>
  File "/Users/jtang/develop/bcfg2/src/lib/Bcfg2/Client/__init__.py", line 458, in load_tools
    self.tools.append(tool(self.config))
TypeError: __init__() takes exactly 4 arguments (2 given)
Loaded tool drivers:
 Action   Bundle   POSIX    VCS      launchd
The following entries are not handled by any tool:
Package:homebrew:cowsay
```

src/lib/Bcfg2/Client/Tools/HomeBrew.py
stpierre: No, just this ``if`` block.

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,74 @@
...
+class HomeBrew(Bcfg2.Client.Tools.PkgTool):
+    """Homebrew package support."""
+    name = 'HomeBrew'
+    __execs__ = ["/usr/local/bin/brew"]
+    __handles__ = [('Package', 'homebrew')]
+    __req__ = {'Package': ['name', 'version']}
+    pkgtype = 'homebrew'
+    pkgtool = ('/usr/local/bin/brew install %s', ('%s', ['name']))
+
+    def __init__(self, logger, setup, config):
```
stpierre: Looks like you may have based this file on the ``MacPorts`` driver from 1.3.x, not 1.4.x; lots of function signatures have changed. In particular, the ``__init__`` for client tools only takes a single argument now.

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,74 @@
...
+                entry.set('current_version', self.installed[entry.get('name')])
+                return False
+        entry.set('current_exists', 'false')
+        return False
+
+    def Remove(self, packages):
+        """Remove extra packages."""
+        pkg_names = " ".join(p.get('name') for p in packages)
+        self.logger.info("Removing packages: %s" % pkg_names)
+        self.cmd.run(["/usr/local/bin/brew", "uninstall"] + pkg_names)
```
stpierre: ``pkg_names`` shouldn't be a string; this addition will fail, and also it'll take the entire list of names as a single argument, which is not what you want. It should be a list, e.g.:

``pkg_names = [p.get('name') for p in packages]``

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,74 @@
...
+        self.RefreshPackages()
+        self.extra = self.FindExtra()
+
+    def InstallPackage(self, packages, states):
```
stpierre: ``Install*`` no longer takes a second argument in 1.4.x.

stpierre: Actually, ``PkgTool`` doesn't even use ``InstallPackage``; it automatically generates the install command from ``cls.pkgtool``. This entire method can be removed.

src/lib/Bcfg2/Client/Tools/HomeBrew.py
jcftang: Yes, that seems to be the case.

jcftang:
The proposed changes have been made

jcftang:
Oops, I need to cleanup the last commit

solj:
Can you also add this to the 1.4.0pre2 release notes?

solj:
:+1:

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,58 @@
...
+                                 "list", "--versions"]).stdout.splitlines()
+        self.installed = {}
+        for pkg in pkgcache:
+            pkgname, version = pkg.strip().split()
+            self.logger.debug(" pkgname: %s version: %s" % (pkgname, version))
+            self.installed[pkgname] = version
+
+    def VerifyPackage(self, entry, _):
+        """Verify Package status for entry."""
+        self.logger.info("VerifyPackage: %s : %s" % (entry.get('name'),
```
stpierre: Log this at debug.

src/lib/Bcfg2/Client/Tools/HomeBrew.py
```diff
@@ -0,0 +1,58 @@
...
+                                                     entry.get('version')))
+        if 'version' not in entry.attrib:
+            self.logger.info("Cannot verify unversioned package %s" %
+                             entry.attrib['name'])
+            return False
```
stpierre: This if block is unreachable.

stpierre:
Github seems to have eaten a couple of my older comments -- maybe it got confused in the rebase.

PR #261, opened 2015-02-16
Fix Decisions Plugin

AlexanderS:
Currently the Decisions plugin does not work because it does not monitors the files. The first patch add the monitoring and the second patch fixes the pseudo filemonitor for not generating the initial exists events, if the file does not exists.


stpierre:
Test errors are unrelated; looks like a new version of pylint. I'll fix them tomorrow.

The first commit makes sense to me, but I want to think a little more about the second. It seems right, but I have to wonder why it wasn't already like that.

solj:
:+1:

stpierre:
You should be able to rebase this against master and get the tests to pass now.

Todo item for me (or anyone else who's a masochist): unit tests for the FAM interface and backends.

PR #263, opened 2015-02-19
Fix extra packages

AlexanderS:
The handling of the reporting of extra packages was a bit strange. The version of the extra package was reported as `target_version` in the reporting. Only yum/rpm packages with a nested `<Instance>` tag parsed the version as current_version.

### DIFF
--- a/src/lib/Bcfg2/Reporting/Storage/DjangoORM.py
+++ b/src/lib/Bcfg2/Reporting/Storage/DjangoORM.py
@@ -109,8 +109,8 @@ class DjangoORM(StorageBase):
         # extra entries are a bit different.  They can have Instance
         # objects
         if not act_dict['target_version']:
-            for instance in entry.findall("Instance"):
-                # FIXME - this probably only works for rpms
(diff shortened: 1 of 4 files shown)


src/lib/Bcfg2/Reporting/Storage/DjangoORM.py
```diff
@@ -109,8 +109,8 @@ def _import_Package(self, entry, state):
         # extra entries are a bit different.  They can have Instance
         # objects
         if not act_dict['target_version']:
-            for instance in entry.findall("Instance"):
-                # FIXME - this probably only works for rpms
+            instance = entry.find("Instance")
```
solj: Does changing findall to find affect the result here?

AlexanderS: Before the patch, there was a `return` in the for loop, so only the first result was handled. Now `find()` makes it at least more obvious that only one element is handled.

stpierre:
+1

solj:
Merged, thanks!

PR #264, opened 2015-02-19
Client/Tools/Pkgng: No auto-update on pkg-search.

AlexanderS:
By default pkg search will update the local copy of the repository
catalogue from remote. We are executing a lot of searches and do
not want to update the catalogue every time.


solj:
Merged, thanks!

PR #265, opened 2015-02-23
Client/Tools/APT: Fix regression of ee4111fc

AlexanderS:
In ee4111fc we removed the wrong continue (the one after the if/else block
for newapi) and now we get an error that 'auto' is not an available version.


solj:
Merged, thanks!

PR #266, opened 2015-02-24
Server/Plugins/Metadata: Reject passwd clients, if auth_type is cert

AlexanderS:
If the auth type (either globally or specific for the client) is set to cert auth,
we have to reject clients that does not provide a cert.


solj:
Merged, thanks!

PR #267, opened 2015-02-24
Fix pylint errors

stpierre:
This also pins pylint to <= 0.28 so we don't have to keep playing
whack-a-mole with it.

Also removes unnecessary suppression of apt warnings. This is no
longer necessary in 12.04, so should be safe to remove. If you're on
Ubuntu < 12.04, upgrade for heaven's sake.

### DIFF
--- a/src/lib/Bcfg2/Client/Tools/APT.py
+++ b/src/lib/Bcfg2/Client/Tools/APT.py
@@ -1,12 +1,10 @@
 """This is the Bcfg2 support for apt-get."""
 
-# suppress apt API warnings
-import warnings
-warnings.filterwarnings("ignore", "apt API not stable yet",
-                        FutureWarning)
-import apt.cache
 import os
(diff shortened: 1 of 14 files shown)


solj:
Hmm, 0.28 was released in 2013. Will we be stuck doing 0.28 indefinitely? Or perhaps this more of a moving target issue wrt travis-ci changing the pylint version on a whim? If we were to make the code pass pylint tests for 1.4.1, would they also pass for < 1.4.1?

stpierre:
I'm not sure why we're just running into these errors now, then -- that's kinda strange.

Pylint actually does a horrible job of backwards compatibility. (And forwards, for that matter.) I should probably update to 1.4 some day, but it won't be terribly trivial.

stpierre:
I'm going to see if we can get pylint 1.4 working for master, and leave maint at 0.28.

PR #268, opened 2015-02-24
Update to patterns used in chkconfig module

gordonmessmer:


src/lib/Bcfg2/Client/Tools/Systemd.py
```diff
@@ -26,31 +26,63 @@ def get_svc_name(self, service):
     def get_svc_command(self, service, action):
         return "/bin/systemctl %s %s" % (action, self.get_svc_name(service))
 
+    def verify_bootstatus(self, entry, bootstatus):
+        """Verify bootstatus for entry."""
+        cmd = self.get_svc_command(entry, 'is-enabled')
+        rv = self.cmd.run(cmd)
+
+        if rv.stdout.strip() == 'enabled':
```
stpierre: I'd prefer using the return value so that we don't have to worry about locales or whatever changing the output of ``systemctl``. E.g.:

    if self.cmd.run(self.get_svc_command(entry, 'is-enabled')):
        return bootstatus == 'on'
    else:
        return bootstatus == 'off'

src/lib/Bcfg2/Client/Tools/Systemd.py
```diff
@@ -26,31 +26,63 @@ def get_svc_name(self, service):
...
+        rv = self.cmd.run(cmd)
+        if rv.stdout.strip() in ('ActiveState=active', 'ActiveState=activating',
+                                 'ActiveState=reloading'):
             entry.set('current_status', 'on')
-            return entry.get('status') == 'on'
+            return entry.get('status') == 'on' and current_bootstatus
         else:
             entry.set('current_status', 'off')
-            return entry.get('status') == 'off'
+            return entry.get('status') == 'off' and current_bootstatus
```
stpierre: Both sections of this if-else are substantially similar. I'd prefer something like:

    if ...:
        current_status = 'on'
    else:
        current_status = 'off'
    entry.set('current_status', current_status)
    return entry.get('status') == current_status and current_bootstatus

src/lib/Bcfg2/Client/Tools/Systemd.py
```diff
@@ -26,31 +26,63 @@ def get_svc_name(self, service):
     def get_svc_command(self, service, action):
         return "/bin/systemctl %s %s" % (action, self.get_svc_name(service))
 
+    def verify_bootstatus(self, entry, bootstatus):
```
stpierre: The second argument here is unnecessary since the bootstatus can be determined from the entry alone. It might provide some slight micro-optimization, but I'd rather not have the requirement to duplicate data in the function call.

src/lib/Bcfg2/Client/Tools/Systemd.py
```diff
@@ -26,31 +26,63 @@ def get_svc_name(self, service):
...
+            return rv
+        elif Bcfg2.Options.setup.servicemode == 'build':
+            # 'build' means we attempt to stop all services started
+            if entry.get('current_status') == 'on':
+                rv &= self.cmd.run(self.get_svc_command(entry, 'stop')).success
+        else:
+            if entry.get('status') == 'on':
+                rv &= self.cmd.run(self.get_svc_command(entry, 'start')).success
+            else:
+                rv &= self.cmd.run(self.get_svc_command(entry, 'stop')).success
```
stpierre: With the way this is restructured I'm not sure it makes sense to compose ``rv`` like this any more. The code might be cleaner if you did something like:

    if bootstatus == 'on':
        rv = self.cmd.run(self.get_svc_command(entry, 'enable')).success
    else:
        rv = self.cmd.run(self.get_svc_command(entry, 'disable')).success
    
    if not rv or Bcfg2.Options.setup.servicemode == 'disabled':
        return rv
    
    if Bcfg2.Options.setup.servicemode == 'build':
        if entry.get('current_status') == 'on':
            return self.cmd.run(self.get_svc_command(entry, 'stop')).success
        else:
            return True
    else:
        if entry.get('status') == 'on':
            return self.cmd.run(self.get_svc_command(entry, 'start')).success
        else:
            return self.cmd.run(self.get_svc_command(entry, 'stop')).success

Although it's kinda messy either way. :(

stpierre:
The commit message ("chkconfig module") doesn't appear to match the change.

src/lib/Bcfg2/Client/Tools/Systemd.py
gordonmessmer: Changed, but now current_status is a string representing the running status of a service, while current_bootstatus is a boolean representing whether the current boot status matches the target boot status.  Which is kind of why the chkconfig module is so hard to read and understand.  I'm going to look at that for a while.

src/lib/Bcfg2/Client/Tools/Systemd.py
gordonmessmer: Sure.  It is the way it is only in order to work more like the Chkconfig module.  If consistency doesn't matter, my opinion is that the function itself is unnecessary.  We've already reduce it to just four lines of functional code, where it's nine lines of code for the function and call.  It's only used once, so moving it into a function kinda makes the code *more* messy rather than less.  If you're not opposed, I'll just move that back into VerifyService.

src/lib/Bcfg2/Client/Tools/Systemd.py
AlexanderS: Maybe something like that for the second part:

```
command = 'stop'
if Bcfg2.Options.setup.servicemode == 'build':
    if entry.get('current_status') == 'off':
        return True
elif entry.get('status') == 'on':
    command = 'start'

return self.cmd.run(self.get_svc_command(entry, command)).success
```

src/lib/Bcfg2/Client/Tools/Systemd.py
stpierre: ``verify_bootstatus`` isn't part of the public interface of ``SvcTool``, so consistency doesn't matter in this case. What you've described makes perfect sense to me.

src/lib/Bcfg2/Client/Tools/Systemd.py
gordonmessmer: That's a very compact implementation, but not very readable, which I treat as a priority for code that's not especially performance sensitive.  The current commit is, I think, highly readable and internally consistent.

gordonmessmer:
Is there any other work you'd like to see done on this before merging?

stpierre:
Can you fix the commit message? (s/chkconfig/systemd/)

gordonmessmer:
Is it possible to change commit messages in git?

What I meant by that message was that I didn't know where to find any proper documentation on the correct structure for a svctool, so I updated this one to more closely follow the implementation present in chkconfig.

stpierre:
Yes, you can amend it with ``git rebase -i`` and then force push to your remote (``git push -f``).

gordonmessmer:
Thanks, Chris.  Updated the commit message.

stpierre:
+1

solj:
:+1:

PR #269, opened 2015-03-04
Update to Pylint 1.4

stpierre:


solj:
Not only does this look exceptional, I learned a few pythonisms while reading the changes. Looks to me like the failures are due to a pylint 1.4/python 2.6 compatibility. :+1:

PR #270, opened 2015-03-16
Packages/Yum.py: Fix dependency resolution logic

solj:
This commit helps the internal YUM resolver to choose the latest version
of a package to generate dependencies. Previously, we were simply
iterating through the file so that the last instance listed always won
(even if that instance was an older version with differing
dependencies).

### DIFF
--- a/src/lib/Bcfg2/Server/Plugins/Packages/Yum.py
+++ b/src/lib/Bcfg2/Server/Plugins/Packages/Yum.py
@@ -1169,6 +1169,94 @@ class YumSource(Source):
                 self.file_to_arch[self.escape_url(fullurl)] = arch
         return urls
 
+    # pylint: disable=R0911,R0912
+    # disabling the pylint errors above because we are interesting in
+    # replicating the flow of the RPM code.
+    def _compare_rpm_versions(self, str1, str2):
+        """ Compare RPM versions.
+
+        This is an attempt to reimplement RPM's rpmvercmp method in python.
+
+        :param str1: package 1 version string
+        :param str2: package 2 version string
+        :return: 1 - str1 is newer than str2
+                 0 - str1 and str2 are the same version
+                -1 - str2 is newer than str1"""
+        if str1 == str2:
+            return 0
+
+        front_strip_re = re.compile('^[^A-Za-z0-9~]+')
+        risdigit = re.compile('(^[0-9]+)')
+        risalpha = re.compile('(^[A-Za-z])')
+        lzeroes = re.compile('^0+')
+
+        while len(str1) > 0 or len(str2) > 0:
+            str1 = front_strip_re.sub('', str1)
+            str2 = front_strip_re.sub('', str2)
+
+            if len(str1) == 0 or len(str2) == 0:
+                break
+
+            # handle the tilde separator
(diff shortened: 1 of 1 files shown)


src/lib/Bcfg2/Server/Plugins/Packages/Yum.py
```diff
@@ -1240,10 +1240,20 @@ def parse_primary(self, data, arch):
...
         for pkg in data.getchildren():
             if not pkg.tag.endswith('package'):
                 continue
             pkgname = pkg.find(XP + 'name').text
+            vtag = pkg.find(XP + 'version')
+            version = "%s%s-%s" % (vtag.get('epoch'), vtag.get('ver'),
+                                   vtag.get('rel'))
+            if pkgname in self.packages[arch]:
+                # skip if version older than a previous version
+                if version < versionmap[pkgname]:
```
mattikus: Does this do a lexicographical `<` instead of an integer based one? Would it make more sense to check the versions explicitly rather than implicitly by doing the string build?

fennm:
I agree with Matt, a lexicographical ordering isn't going to work properly for versions.  As a simple example:

    >>> sorted(["2.6","2.12","2.1"])
    ['2.1', '2.12', '2.6']

RPM has a function called `labelCompare` which can be used as a comparator.  It takes two 3-tuples (epoch, version, release) and does a comparison:

    >>> import rpm
    >>> rpm.labelCompare(("1", "2.6", "13"), ("1","2.12","14"))
    -1
    >>> rpm.labelCompare(("1", "2.61", "13"), ("1","2.12","14"))
    1

solj:
Right, I'm going to reimplement using the algorithm from RPM.

solj:
Okay, version check redone according to RPM's algorithm.

http://rpm.org/gitweb?p=rpm.git;a=blob;f=lib/rpmvercmp.c;h=b3d08faa4a31354d821cb259106b182410e452fb;hb=HEAD

src/lib/Bcfg2/Server/Plugins/Packages/Yum.py
```diff
@@ -1237,13 +1325,24 @@ def parse_primary(self, data, arch):
...
         if arch not in self.provides:
-            self.provides[arch] = dict()
+            self.provides[arch] = {}
+        versionmap = {}
         for pkg in data.getchildren():
             if not pkg.tag.endswith('package'):
                 continue
             pkgname = pkg.find(XP + 'name').text
+            vtag = pkg.find(XP + 'version')
+            version = "%s%s-%s" % (vtag.get('epoch'), vtag.get('ver'),
```
fennm: Concatenating the epoch to the front of the version might not always work, for example:

package A:
epoch: 1
version: 11.0
release: 1
str: 111.0-1

package B:
epoch: 2
version: 1.0
release: 1
str: 21.0-1

Package B should be a higher version due to the higher epoch, but this string format would cause A to be higher.

More generally, I believe that each component of the version should be handled separately as in the RPM code.

solj:
@fennm switched to compare each part separately.

fennm:
+1 looks good to me

mattikus:
:+1: LGTM

PR #273, opened 2015-03-19
compat: use eval instead of ast.literal_eval on python 2.4

fennm:


solj:
:+1:

doc/development/compat.txt
```diff
@@ -129,6 +129,8 @@ below, since some of these implementations may be feature-incomplete.
 +----------------+--------------------------------+--------------------------------------------+
 | MutableMapping | :class:`UserDict.DictMixin`    | :class:`collections.MutableMapping` (2.6+) |
 +----------------+--------------------------------+--------------------------------------------+
+| literal_eval   | :func:`ast.literal_eval`       | :func:`eval`                               |
```
stpierre: Shouldn't these be the other way around?

fennm: Good catch.  I fixed that, and also changed the header to be more clear

stpierre:
+1

solj:
I think ast.literal_eval is "technically" only 2.6+.

solj:
Merged, thanks!

fennm:
Thanks!

PR #274, opened 2015-03-25
Server/Plugins/Packages: Fix _init_attributes position.

AlexanderS:
``_init_attributes`` should be called after all properties of the ``Source`` class
are initialized (so that ``_init_attributes`` could overwrite some of it).
The ``Yum`` class initializes ``self.deps`` with a different default entry, that
should not be reset by ``__init__`` of ``Source`` afterwards.


solj:
Merged, thanks!

PR #276, opened 2015-04-15
handle filesystem secontexts properly for contextless filesystems

### LINKED ISSUE #275: Client marks files incorrect on file systems that don't support file labels
The client marks files as incorrect if the filesystem doesn't support file labels with the following errors during runtime.  
```
POSIX: Failed to restore SELinux context for /boot/efi/EFI/Microsoft/Boot/refind
.conf: [Errno 95] Operation not supported
POSIX: Backup of /var/caen/hometeam/.k5login saved to /var/cache/bcfg2/backup/_v
ar_caen_hometeam_.k5login_2015-04-13T11:46:32.531324
POSIX: Backup of /boot/efi/EFI/refind/background.png saved to /var/cache/bcfg2/b
ackup/_boot_efi_EFI_refind_background.png_2015-04-13T11:46:32.535159
POSIX: Failed to restore SELinux context for /boot/efi/EFI/refind/background.png
HlA7QN: [Errno 95] Operation not supported
POSIX: Failed to restore SELinux context for /boot/efi/EFI/refind/background.png
: [Errno 95] Operation not supported
POSIX: Backup of /boot/efi/EFI/refind/manual.conf saved to /var/cache/bcfseg2/back
up/_boot_efi_EFI_refind_manual.conf_2015-04-13T11:46:32.583443
POSIX: Failed to restore SELinux context for /boot/efi/EFI/refind/manual.confDvU
q4i: [Errno 95] Operation not supported

Incorrect entries:      14
Path:file:/boot/efi/EFI/Microsoft/Boot/refind.conf
Path:file:/boot/efi/EFI/refind/background.png
Path:file:/boot/efi/EFI/refind/manual.conf
```
Looking at the code in maint it seems that there's a couple obvious ways to make the client not mark the files as incorrect.  

Both involve an alteration to _set_secontext in POSIX/base.py   

Option 1 would be to check if the OSError thrown around lines 292,300 to check if the errorno is 95, and then return true.  
Option 2 would be to detect if the file context contains dosfs_t or fusefs_t and not try to set a label at all.

gaurdro:
- adds 'secontext_ignore' under POSIX in the configuration file
- short circuits on filesystems that are known not to support file labels
- defaults to filesystems that have a genfs command in selinux reference policy
- checks for Operation not supported while setting a file label
- fixes #275

### DIFF
--- a/src/lib/Bcfg2/Client/Tools/POSIX/base.py
+++ b/src/lib/Bcfg2/Client/Tools/POSIX/base.py
@@ -6,6 +6,7 @@ import pwd
 import grp
 import stat
 import copy
+import errno
 import shutil
 import Bcfg2.Client.Tools
 import Bcfg2.Client.XML
@@ -272,7 +273,7 @@ class POSIXTool(Bcfg2.Client.Tools.Tool):
             rv &= self._apply_acl(defacl, path, posix1e.ACL_TYPE_DEFAULT)
         return rv
 
-    def _set_secontext(self, entry, path=None):
+    def _set_secontext(self, entry, path=None):  # pylint: disable=R0911
         """ set the SELinux context of the file on disk according to the
         config"""
         if not HAS_SELINUX:
@@ -284,25 +285,28 @@ class POSIXTool(Bcfg2.Client.Tools.Tool):
         if not context:
             # no context listed
             return True
-
-        if context == '__default__':
-            try:
+        secontext = selinux.lgetfilecon(path)[1].split(":")[2]
+        if secontext in self.setup["posix_secontext_ignore"]:
+            return True
+        try:
+            if context == '__default__':
                 selinux.restorecon(path)
-                rv = True
-            except OSError:
-                err = sys.exc_info()[1]
-                self.logger.error("POSIX: Failed to restore SELinux context "
-                                  "for %s: %s" % (path, err))
-                rv = False
-        else:
-            try:
-                rv = selinux.lsetfilecon(path, context) == 0
-            except OSError:
-                err = sys.exc_info()[1]
-                self.logger.error("POSIX: Failed to restore SELinux context "
-                                  "for %s: %s" % (path, err))
-                rv = False
-        return rv
+                return True
+            else:
+                return selinux.lsetfilecon(path, context) == 0
+        except OSError:
+            err = sys.exc_info()[1]
+            if err.errno == errno.EOPNOTSUPP:
+                # Operation not supported
+                if context != '__default__':
+                    self.logger.debug("POSIX: Failed to set SELinux context "
+                                      "for %s: %s" % (path, err))
+                    return False
+                return True
+            err = sys.exc_info()[1]
+            self.logger.error("POSIX: Failed to set or restore SELinux "
+                              "context for %s: %s" % (path, err))
+            return False
 
     def _norm_gid(self, gid):
         """ This takes a group name or gid and returns the
--- a/src/lib/Bcfg2/Options.py
+++ b/src/lib/Bcfg2/Options.py
@@ -1115,6 +1115,18 @@ CLIENT_POSIX_GID_BLACKLIST = \
            default=[],
            cf=('POSIXUsers', 'gid_blacklist'),
            cook=list_split)
+CLIENT_POSIX_SECONTEXT_IGNORE = \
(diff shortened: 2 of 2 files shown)


src/lib/Bcfg2/Client/Tools/POSIX/base.py
```diff
@@ -280,24 +280,23 @@ def _set_secontext(self, entry, path=None):
...
-            try:
+            else:
                 rv = selinux.lsetfilecon(path, context) == 0
-            except OSError:
+        except OSError, e:
+                if e.errno == 95:
+                    # Operation not supported
+                    return True
                 err = sys.exc_info()[1]
                 self.logger.error("POSIX: Failed to restore SELinux context "
```
stpierre: It looks like there was a bug in the original code; I'm guessing there were two separate try-except blocks so that the error message could be different in each case. I like the flow of your code better, though, but let's change this error message to "Failed to set or restore..."

src/lib/Bcfg2/Client/Tools/POSIX/base.py
```diff
@@ -280,24 +280,23 @@ def _set_secontext(self, entry, path=None):
...
-                err = sys.exc_info()[1]
-                self.logger.error("POSIX: Failed to restore SELinux context "
-                                  "for %s: %s" % (path, err))
-                rv = False
-        else:
-            try:
+            else:
                 rv = selinux.lsetfilecon(path, context) == 0
-            except OSError:
+        except OSError, e:
```
stpierre: This syntax won't work for all versions of Python that we need to support. You'll need to do:

    except OSError:
        err = sys.exc_info()[1]
        if err.errno == 95:
            ...

src/lib/Bcfg2/Client/Tools/POSIX/base.py
```diff
@@ -280,24 +280,23 @@ def _set_secontext(self, entry, path=None):
...
+                if e.errno == 95:
+                    # Operation not supported
+                    return True
```
stpierre: If someone has explicitly set a context on a file (i.e., context != __default__), and that file is on a selinux-less filesystem, then we should at least log a warning, and possibly return False here.

stpierre:
This looks like a good approach, just needs some cleanup and unit test updates.

solj:
:+1:

src/lib/Bcfg2/Client/Tools/POSIX/base.py
```diff
@@ -280,28 +281,32 @@ def _set_secontext(self, entry, path=None):
...
-                rv = False
+        except OSError:
+            err = sys.exc_info()[1]
+            if err.errno == errno.EOPNOTSUPP:
+                # Operation not supported
+                if context != '__default__':
+                    self.logger.warning("POSIX: Failed to set SELinux context "
+                                        "for %s: %s" % (path, err))
+                    return False
+                return True
```
stpierre: Let's log this condition at a debug level. It shouldn't be concerning, but it'd be nice to have.

src/lib/Bcfg2/Client/Tools/POSIX/base.py
```diff
@@ -280,28 +281,32 @@ def _set_secontext(self, entry, path=None):
...
+            err = sys.exc_info()[1]
+            self.logger.error("POSIX: Failed to set or restore SELinux "
+                              "context for %s: %s" % (path, err))
+            rv = False
         return rv
```
stpierre: I know I'm tacking on semi-related stuff now, but while we're refactoring this function can we get rid of this ``rv`` silliness that I put in and just return what we want to return when we want to return it? I have a feeling I just did a bit too much uncritical copypasta from one of the other functions that actually uses and modifies the return value before returning it.

stpierre:
Looking very, very good.

gaurdro:
I moved checking the known contextless file contexts below checking if there is any context to be applied, since it's likely a slower check because it has to read from disk.

gaurdro:
womp, womp.  

```
-------------------- >> begin captured stdout << ---------------------
************* Module Bcfg2.Client.Tools.POSIX.base
R0911:276,4:POSIXTool._set_secontext: Too many return statements (8/6)
```

stpierre:
Hahaha, you can tell pylint to ignore that with:

    # pylint: disable=R0911

inline with the function declaration.

PR #277, opened 2015-04-18
Server: Ignore directories containing a .bcfg2-ignore file

AlexanderS:
The DirectoryBacked helper now does not recurse into directories that
contains a .bcfg2-ignore file. This makes it possible to ignore some
directories for most plugins (Bundler, Defaults, Pkgmgr, Properties,
PuppetENC, TemplateHelper, Trigger). You can store for example a python
module used by a TemplateHelper in the same directory, without getting
strange error messages while TemplateHelper is trying to import each
single file of this module.

### DIFF
--- a/doc/releases/1.4.0pre2.txt
+++ b/doc/releases/1.4.0pre2.txt
@@ -31,6 +31,9 @@ backwards-incompatible user-facing changes
   This fixes potentially long client runs when comparing files that have
   diverged significantly.
 
+* Ignore directories containing a .bcfg2-ignore file in various plugins
+  (Bundler, Defaults, Pkgmgr, Properties, PuppetENC, TemplateHelper, Trigger).
+
 Thanks
 ------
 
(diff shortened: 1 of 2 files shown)


stpierre:
I like this, and it's simple enough. I wonder, though, if it wouldn't be better to do something like ``.gitignore``, where the file contains a list of globs to ignore. Ideally, subdirectories would inherit parent directories' ignore list, although that could be difficult.

That way we could also ignore files, not just directories. Basically, it'd be a better, more granular version of the server.ignore_files configuration directive.

Thoughts?

mattikus:
@stpierre Your suggestion sounds wonderful to me. :+1:

AlexanderS:
For a `.bcfg2-ignore` file like a `.gitignore` file, it would be better to do the processing inside the FileMonitor plugin (like server.ignore_files). But the FileMonitor plugin may not have the full path for an event. At least the Inotify plugin already caches the path for the handler. Should I pull the path caching into the FileMonitor base class and the add dynamic ignores there?

stpierre:
Yes, I think so -- at least, if it's possible.

Thanks!

PR #278, opened 2015-05-09
bcfg2-admin: Add schemamigration command from south

AlexanderS:
This adds `schemamigration` as a new `bcfg2-admin` subcommand, so while developing one could update the south migrations easily.

### DIFF
--- a/src/lib/Bcfg2/Server/Admin.py
+++ b/src/lib/Bcfg2/Server/Admin.py
@@ -877,6 +877,7 @@ if HAS_DJANGO:
         Django management system """
         command = None
         args = []
+        kwargs = {}
 
         def run(self, _):
             '''Call a django command'''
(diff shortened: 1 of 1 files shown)


stpierre:
+1

AlexanderS:
Maybe we also want a manual way to execute the migrations. It would be easy like this:

```
    class Migrate(_DjangoProxyCmd):
        """ Call the south 'migrate' command on the database """
```

solj:
Does this subcommand also exist in the django 1.7 migration framework?

AlexanderS:
Do we want to drop south in favor of the migration framework of django 1.7? According to the [docs](https://docs.djangoproject.com/en/1.8/topics/migrations/) django-1.7 uses makemigrations instead.

solj:
@AlexanderS I was considering it unless there's some reason we need to keep south around. It seems like our migrations are fairly straightforward and if we can reduce our dependency footprint, it'd be a win.

AlexanderS:
@solj Currently the web interface does not work with django-1.7, or does it? Maybe we could just commit this in the current form (maybe with the "migrate" subcommand) and do the transition to the django migration framework with the support for django-1.7? I am sure that there are some more things to change...

AlexanderS:
Now also for `makemigrations` with compatibility with django-1.7.

solj:
Merged, thanks!

PR #279, opened 2015-05-10
Add some flags for reporting

AlexanderS:
This records the `--dry-run` and `--only-important` flags in the reporting database and
adds the possibility to exclude the dry-run and only-important runs in the `bcfg2-reports`
output.

### DIFF
--- a/src/lib/Bcfg2/Client/__init__.py
+++ b/src/lib/Bcfg2/Client/__init__.py
@@ -929,6 +929,11 @@ class Client(object):
                                'Statistics', total=str(len(states)),
                                version='2.0',
                                revision=self.config.get('revision', '-1'))
+        flags = XML.SubElement(stats, "Flags")
+        XML.SubElement(flags, "Flag", name="dry_run",
+                       value=str(Bcfg2.Options.setup.dry_run))
+        XML.SubElement(flags, "Flag", name="only_important",
+                       value=str(Bcfg2.Options.setup.only_important))
         good_entries = [key for key, val in list(states.items()) if val]
         good = len(good_entries)
         stats.set('good', str(good))
--- a/src/lib/Bcfg2/Reporting/Reports.py
+++ b/src/lib/Bcfg2/Reporting/Reports.py
@@ -7,57 +7,38 @@ import datetime
 import Bcfg2.DBSettings
 
 
-def hosts_by_entry_type(clients, etype, entryspec):
-    result = []
-    for entry in entryspec:
-        for client in clients:
-            items = getattr(client.current_interaction, etype)()
-            for item in items:
-                if (item.entry_type == entry[0] and
-                    item.name == entry[1]):
-                    result.append(client)
-    return result
-
-
-def print_fields(fields, client, fmt, extra=None):
-    """ Prints the fields specified in fields of client, max_name
-    specifies the column width of the name column. """
-    fdata = []
-    if extra is None:
-        extra = dict()
-    for field in fields:
-        if field == 'time':
-            fdata.append(str(client.current_interaction.timestamp))
-        elif field == 'state':
-            if client.current_interaction.isclean():
-                fdata.append("clean")
-            else:
-                fdata.append("dirty")
-        elif field == 'total':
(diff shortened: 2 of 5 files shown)


src/lib/Bcfg2/Client/__init__.py
```diff
@@ -925,10 +925,14 @@ def GenerateStats(self):
...
         feedback = XML.Element("upload-statistics")
-        stats = XML.SubElement(feedback,
-                               'Statistics', total=str(len(states)),
-                               version='2.0',
-                               revision=self.config.get('revision', '-1'))
+        stats = XML.SubElement(
+            feedback,
+            'Statistics',
+            total=str(len(states)),
+            version='2.0',
```
solj: It seems like we're never going to use this 'version' attribute. Perhaps we should get rid of it?

solj:
Does this allow older clients to work as they currently do while adding the functionality for new clients that support it?

src/lib/Bcfg2/Reporting/Reports.py
```diff
@@ -132,7 +117,42 @@ def get_clients(self):
         from Bcfg2.Reporting.models import Client
         return Client.objects.exclude(current_interaction__isnull=True)
 
-    def display(self, result, fields, extra=None):
+    def _print_fields(self, setup, fields, client, fmt, extra=None):
```
stpierre: In 1.4 there's no need to pass the 'setup' object around, since it can be accessed as a module-level ``Bcfg2.Options.setup``.

src/lib/Bcfg2/Client/__init__.py
```diff
@@ -925,10 +925,14 @@ def GenerateStats(self):
...
+            revision=self.config.get('revision', '-1'),
+            dry_run=str(Bcfg2.Options.setup.dry_run),
+            only_important=str(Bcfg2.Options.setup.only_important))
```
stpierre: I think i'd like this more if, instead of adding two new top-level tags for these flags, you added a new section for flags so that we can track other flags in the future. (Other flags, after all, can affect which entries are processed.) E.g.:

    stats = XML.SubElement(feedback,
                           'Statistics',
                           total=str(len(states)),
                           version='2.0',
                           revision=self.config.get('revision', '-1'))
    XML.SubElement(stats, "Flags",
                   dry_run=str(Bcfg2.Options.setup.dry_run),
                   only_important=str(Bcfg2.Options.setup.only_important))

I'm not sure right off how that'd affect the statistics parsing, though.

stpierre:
I <3 this feature.

src/lib/Bcfg2/Client/__init__.py
AlexanderS: Maybe even:

```python
flags = XML.SubElement(stats, "Flags")
XML.SubElement(flags, "Flag", name="dry_run",
               value=str(Bcfg2.Options.setup.dry_run))
XML.SubElement(flags, "Flag", name="only_important", 
               value=str(Bcfg2.Options.setup.only_important))
```

Or do you think that is over complicated?

AlexanderS:
@solj: Parsing defaults the values to ```False``` if they do not exists. So it should work with older clients.

src/lib/Bcfg2/Client/__init__.py
stpierre: That makes sense to me.

solj:
@AlexanderS Awesome, looks good!

src/lib/Bcfg2/Reporting/Reports.py
AlexanderS: Reading the doc string of Subcommand.run() (https://github.com/Bcfg2/bcfg2/blob/master/src/lib/Bcfg2/Options/Subcommands.py#L110-120) it seems, that you should not use `Bcfg2.Options.setup` in a Subcommand.

AlexanderS:
I just updated the format of the flags in the statistics. (I do not know, if github send notifications on changed commits.)

stpierre:
LGTM.

solj:
Merged, thanks!

PR #280, opened 2015-05-10
Some small fixes for SSHbase

AlexanderS:
This fixes two small things in SSHbase:

* The logging call is a regression from 3d06f3112 (and therefor the error is not found in miant).
* The exception types are also wrong in maint, but it does not break things there, because in maint there is a wildcard except around.


solj:
:+1: LGTM.

PR #281, opened 2015-05-22

feuri:
This makes Bcfg2-web work with django 1.7 (and probably also 1.8).
I have not yet checked whether all 'bcfg2-admin reports' commands still work and running manage.py manually currently outputs that django 1.6 has introduced a new test runner and that tests might not be running as expected.


solj:
@feuri I think we would prefer to apply this to the master branch. The maint branch is in bugfix mode and this looks like a pretty invasive change. Could you create a patch for master?

PR #282, opened 2015-06-03

feuri:
- reports.wsgi uses get_wsgi_application() now
  - old south-based migrations have been moved
  - manage.py has been updated

Same as #281 but now against master


feuri:
```
E1101:1204,12:CLI.__init__: Module 'django' has no 'setup' member
```
django 1.4 doesn't have the django.setup() method, so pylint fails.

solj:
I'm still getting the following with master + your commits. This is with django 1.7.7

```
# bcfg2-server 
Updating database /bcfg2-web/bcfg2.sqlite failed: Apps aren't loaded yet.
service available at https://bcfg2:6789
Django south is required for Reporting
Failed to instantiate plugin <class 'Bcfg2.Server.Plugins.Reporting.Reporting'>
  File "/root/bcfg2/src/lib/Bcfg2/Server/Core.py", line 421, in init_plugin
    self.plugins[plugin.name] = plugin(self)
  File "/root/bcfg2/src/lib/Bcfg2/Server/Plugins/Reporting.py", line 54, in __init__
    raise PluginInitError(msg)
Bcfg2.Server.Plugin.exceptions.PluginInitError: Django south is required for Reporting
```

feuri:
Well, I haven't worked on this PR as I was working on fixing django 1.7 support for bcfg2 1.3.5 first.

AlexanderS:
It seems, that the migrations are not correctly converted into the new format. For example the field `dry_run` of `Interaction` was a migration before, but it seems to be part of the base model now.

feuri:
I've recreated new migrations based on the instructions found [here](https://docs.djangoproject.com/en/1.7/topics/migrations/#upgrading-from-south), but I haven't had a look at whether the new migrations are actually correct.

feuri:
I have had a look at the migrations again and split the migrations originally done in 0007_add_flag_fields_interaction.py from the initial migration. You can have a look at it [here](https://github.com/feuri/bcfg2/tree/split_django_migrations/src/lib/Bcfg2/Reporting/migrations). If it looks good to you, I'll merge the commit into this PR.

AlexanderS:
We should keep the south migrations in `south_migrations` so that they also can be applied. The upgrade process from south to django migrations does not keep the state of the applied migrations. So to upgrade to django migrations, all south migrations should be applied first and then all django migrations should be fake applied.

I think, we have to support south and django migrations in parallel for at least one release.

solj:
@AlexanderS Can we use both at the same time? Do we need to have a migration script that runs the south migrations and then only include the django migrations?

AlexanderS:
Yes it should be possible to support both at the same time. At the moment someone would like to migrate to django migrations all south migrations should be applied and all available django migrations should be fake applied.

solj:
Closing as duplicate of #311

PR #285, opened 2015-07-16
Systemd: Backport the code from master into maint

jsbillings:
Taken directly from master (ref 33e53dde2a85b8783c8e4935868d9c5f50dea440).
Only a minor change, using:

self.setup['servicemode']

instead of:

Bcfg2.Options.setup.servicemode

This change is needed because the Systemd tool in maint is horribly out of date, and doesn't fully manage systemd services.  For example, the VerifyService function didn't check whether services were enabled/disabled, just whether they were running or not.


AlexanderS:
The travis error seems unrelated. It is still the last commit of maint that causes the failure.

solj:
Yep, reran and everything looks good. :+1:

PR #286, opened 2015-07-20
DBSettings: Don't use django, if it's not available

AlexanderS:


stpierre:
+1

solj:
Isn't django required for DBSettings?

stpierre:
The option parsing hooks are called regardless, though, so they should be no-ops if Django isn't available.

Aside: For 1.5 we should seriously look into making more things requirements. This is brittle and a little silly.

solj:
Merged, thanks!

PR #287, opened 2015-07-20
Plugins/Cfg: Don't depend on jinja2

AlexanderS:
If jinja2 is not available, we cannot use some classes. So the access to
these classes have to be dependent on the available of jinja2.

I do not know, if it is good to have an if around assignments on class level. As an alternative we could create two dummy classes (`def Environment(object): pass` or something like that) if jinja2 is not available.


stpierre:
I wish there was a prettier way to do this, but I can't think of one. +1

PR #288, opened 2015-07-20
Fix some issues with new Options parser in tools/

AlexanderS:
Hi,

the migration to the new parser for Options has introduced some small issues in tools/. The new parser does not read the Plugin names from the config but instantiates the full Plugin classes. So we need `__name__` to access the raw name.

Another issue is, that the components should be specified as list, even if it is only one.

Alex


solj:
:+1:

PR #289, opened 2015-07-21
PrioDir: Only set text content, if something is available

AlexanderS:
This is used for the Rules and Defaults plugins. Without this patch all text
content that was there before (for example file content bound by Cfg) is lost,
for all entries specified in Defaults.


stpierre:
+1

solj:
Merged, thanks!

PR #290, opened 2015-07-21
Lint/MergeFiles: Ignore binary files

AlexanderS:
Ignore files with binary content, because SequenceMatcher seems to have problems
and sometimes detect files with different content as identically.

### DIFF
--- a/src/lib/Bcfg2/Client/Tools/POSIX/File.py
+++ b/src/lib/Bcfg2/Client/Tools/POSIX/File.py
@@ -8,6 +8,7 @@ import tempfile
 import Bcfg2.Options
 from Bcfg2.Client.Tools.POSIX.base import POSIXTool
 from Bcfg2.Compat import unicode, b64encode, b64decode  # pylint: disable=W0622
+import Bcfg2.Utils
 
 
 class POSIXFile(POSIXTool):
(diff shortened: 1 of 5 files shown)


solj:
:+1:

stpierre:
We have a different way to check if a string is binary here: https://github.com/Bcfg2/bcfg2/blob/maint/src%2Flib%2FBcfg2%2FClient%2FTools%2FPOSIX%2FFile.py#L20-L25 (We negate that check later: https://github.com/Bcfg2/bcfg2/blob/maint/src%2Flib%2FBcfg2%2FClient%2FTools%2FPOSIX%2FFile.py#L184)

Should we use the same approach in both places? Is one better than the other?

AlexanderS:
Oh, the first part of the check is essentially the same, it checks if some bytes are outside the printable range in extended ascii. The check in POSIXFile also checks if the encoding is valid. So maybe we should move the check from POSIXFile into Utils or so and use this version.

AlexanderS:
Done.

stpierre:
LGTM.

AlexanderS:
*grml* The test needs to be fixed.

solj:
Merged, thanks!

PR #292, opened 2015-07-21
Options: Set default values from config, right after adding a new option

AlexanderS:
If adding an option during the "Main Parser Loop" (for example because of loading
a component for bcfg-lint) a value for the option from the config file is simply
ignored.

After adding the option the parser first tries to find the value in the command line,
but cannot find it and set the default value from the source code as option value.
After that the value from the config file is set as new default. But because the
option already is in the Namespace, it does not use the new "default" value from the
config file.

This patch simply sets the default value from the config file, right after the new option
is added to the parser and so the correct value is used afterwards, if the parsed cannot
find the flag on the command line.


solj:
:+1:

PR #294, opened 2015-07-22
Rules: New options replace_name to replace ${name} in attributes

AlexanderS:
If you use the regex feature of Rules/Defaults you may need the real name
of the matched entry in an attribute (for example home of POSIXUser).
You can now enable replace_name for rules or defaults and ${name} will be
replaces in the attribues of the Element before adding them to the target
entry.

This allows you to write something like that in Defaults to assing a default
home directory somewhere else to all users with unset home:

    <POSIXUser name='.*' home='/somewhere/${name}'/>

### DIFF
--- a/doc/server/plugins/generators/rules.txt
+++ b/doc/server/plugins/generators/rules.txt
@@ -512,3 +512,23 @@ you'd have to explicitly specify ``<Service name="bcfg2.*".../>``.
 
 Note that only one Rule can apply to any abstract entry, so you cannot
 specify multiple regexes to match the same rule.
+
+Replacing the name of the Entry in Attributes
+=============================================
+
+If you are using regular expressions to match the abstract configuration
+entries, you may need the concrete name of the entry in some attributes.
+To use this feature, you have to enable it. It is only useful, if used
+together with regex matching. ::
+
+    [rules]
+    regex = yes
+    replace_name = yes
+
+You now can write something like that in your xml file:
+
+.. code-block:: xml
+
+    <POSIXUser name='.*' home='/somewhere/%{name}'/>
+
+``%{name}`` will be correctly replaced with the username for each POSIXUser.
--- a/doc/releases/1.4.0pre2.txt
+++ b/doc/releases/1.4.0pre2.txt
@@ -19,6 +19,7 @@ environments.
 
 * NagiosGen: Add bundles to configuration
 * HomeBrew: Initial add of plugin
+* Rules/Defaults: Add possibility to use name of entry in attributes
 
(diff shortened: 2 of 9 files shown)


doc/server/plugins/generators/rules.txt
```diff
@@ -512,3 +512,23 @@ you'd have to explicitly specify ``<Service name="bcfg2.*".../>``.
 
 Note that only one Rule can apply to any abstract entry, so you cannot
 specify multiple regexes to match the same rule.
+
+Replacing the name of the Entry in Attribues
```
solj: Should be Attributes

doc/server/plugins/generators/rules.txt
```diff
@@ -512,3 +512,23 @@ you'd have to explicitly specify ``<Service name="bcfg2.*".../>``.
...
+============================================
+
+If you are using regular expressions to match the abstract configuration
+entries, you may need the concrete name of the entry in some attributes.
+To use this feature, you have to enable it. It is only usefull, if used
```
solj: another typo "useful"

doc/server/plugins/generators/rules.txt
```diff
@@ -512,3 +512,23 @@ you'd have to explicitly specify ``<Service name="bcfg2.*".../>``.
...
+together with regex matching. ::
+
+    [rules]
+    regex = yes
+    replace_name = yes
+
+You now can write somehing like that in your xml file:
```
solj: typo "something"

solj:
This looks good to me at first glance. Can you also mention the feature in the release notes?

AlexanderS:
Fixed the typos and added a line to the release notes.

solj:
:+1:

stpierre:
This seems like it won't work for (or at least will conflict with) Rules or Defaults files that are Genshi templates, since ``${name}`` will look like a Genshi variable. So depending on the order of evaluation, either a) I can no longer use ``name`` as a variable in my Genshi template; or b) I can't use replace_name in Genshi templates. Either way it's unexpected.

I'd suggest using a different format for the name token. ``%(name)s`` would be easy enough from a code point of view, but I'm open to other suggestions as long as they don't conflict with Genshi templating.

Additionally, I can see the use case for this in Defaults, but not so much in Rules. (I've never liked the Rules plugin, though, so I might be biased.) What's the use case in Rules?

AlexanderS:
I simply replaced the delimiter for `string.Template` with `%`, so that you can use `%name` or `%{name}`. This has the advantage over the classic `%` format strings, that it is more robust and does not fail if extra placeholders are in the string.

PR #295, opened 2015-07-23
Options/Types: Fix database OPTIONS parsing

AlexanderS:
This is mostly from 257eb0c17 and 16d3e04cb to allow for setting arbitrary
database options since some settings require nested dictionaries, etc.

This got lost during merge from maint into master because of the completely
different Options parsing.


solj:
:+1:

PR #296, opened 2015-07-24
Ldap: Cleanup plugin

AlexanderS:
Hi,

first I only wanted to investigate, how we should handle cache expiration with the Ldap plugin, but now it is a complete cleanup ;)

The caching is a bit of a problem. The plugin cannot know if there are changes on the LDAP server, so it expires the full client metadata cache if the configuration changes and at beginning of a client run in aggressive caching mode it expires the metadata for that single client. If you use Ldap data from other clients metadata in a template, you may get outdated values from the last run of that other client, but it is better then generating each client metadata on every client run.

This also changes the interface for the configuration file in the repository, but I think it shouldn't be a problem, because the plugin is marked as experimental and configuration should be easily adjustable.

Alex


doc/server/plugins/grouping/ldap.txt
```diff
@@ -108,143 +120,140 @@ LdapQuery
...
+that list, that should be called automatically and whose results should be added to the
+client metadata.
 
 Known Issues
 ------------
 
 * At this point there is no support for SSL/TLS.
+* This module could not know, if a value changed on the LDAP server. So it could not
+  expire the client metadata cache sanely.
+  If you are using aggressiv caching mode, this plugin will expire the metadata cache
```
solj: aggressive

doc/server/plugins/grouping/ldap.txt
```diff
@@ -108,143 +120,140 @@ LdapQuery
...
+  for a single client at the start of a client run. If you are using LDAP data from
+  another client in a template, you will propably get the cached values from the last
```
solj: probably

doc/releases/1.4.0pre2.txt
```diff
@@ -36,6 +36,15 @@ backwards-incompatible user-facing changes
 
   This allows to set arbitrary options with nested settings.
 
+* The Ldap plugin changed significant. The configuration interface was
```
solj: significantly

doc/releases/1.4.0pre2.txt
```diff
@@ -36,6 +36,15 @@ backwards-incompatible user-facing changes
...
+  simplified and new configuration options for the number of retries and the
+  delay in between were added.
+
+  You have to register your ladp queries in the global list, there is no
```
solj: ldap

solj:
Other than the comments from the grammar nazi, this looks good to me :+1:

AlexanderS:
Fixed typos.

solj:
Merged, thanks!

PR #297, opened 2015-07-24
Travis-CI: Use new container-based infrastructure

AlexanderS:
Travis-ci started to promote the new container-base infrastructure in each build on the legacy systems. So we should try to migrate. I requested to add yum and libaugeas0 to the package whitelist, so it should be possible to install all required apt-packages.

Let's see how much speed-up we will get from this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ matrix:
       addons:
         apt:
           packages: *build_deps_optional
+      script: nosetests --show-skipped testsuite
 
     - python: "2.7"
     - python: "2.7_with_system_site_packages"
@@ -36,6 +37,7 @@ matrix:
       addons:
         apt:
           packages: [*build_deps_optional, python-gamin, python-selinux]
+      script: nosetests --show-skipped testsuite
 
 before_install:
   - testsuite/prepare-python.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ sudo: false
 language: python
 matrix:
   include:
-    - env: PYTHON=2.4
-    - env: PYTHON=2.5
+    - env: PYTHON=2.4 NOSE_EXCLUDE_DIRS=testsuite/Testsrc/Testlib/TestServer
+    - env: PYTHON=2.5 NOSE_EXCLUDE_DIRS=testsuite/Testsrc/Testlib/TestServer
     - python: "2.6"
     - python: "2.7"
     - python: "2.7_with_system_site_packages"
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,6 @@ matrix:
       addons:
         apt:
           packages: *build_deps_optional
-      script: nosetests --show-skipped testsuite
 
     - python: "2.7"
     - python: "2.7_with_system_site_packages"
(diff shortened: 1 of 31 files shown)


solj:
:+1:

stpierre:
We should also see if we can get selinux libs installed (since we currently don't test that), and see if we can test in older Python somehow. That might be scope creep for this PR, though.

AlexanderS:
I think we can download, compile and install older python version. There is a guide for installing custom software: http://docs.travis-ci.com/user/migrating-from-legacy/#How-Do-I-Install-Custom-Software%3F

AlexanderS:
Most stuff is working for python2.4 and python2.5 now. Only register_transformer is not available with a pylint that is compatible with python2.4. Additionally python-daemon (in a compatible version) does not drop permissions and so we cannot use this function.

The only think I do not understand is the stuff with the infinitive recursion while copying some xdata with python2.4.

solj:
I thought we were going to go ahead and require python 2.6 for 1.4.

stpierre:
I thought the plan was to require 2.6 on the server only (I.e., 2.4 still
supported on the client) by 1.5. (I'd be fine doing that in 1.4, for the
record. We can de-Bcfg2.Compat some things.) We need to support 2.4 on the
client side for a while yet, so it'd be nice to be able to run at least the
relevant part of the test suite.

solj:
@stpierre Ah, you're right.

AlexanderS:
I have no idea why pylint does not get the imports...

AlexanderS:
:tada: Working tests with python2.4, python2.5, python2.6 and python2.7.

It was a hassle to get pylint working with older python versions. Don't ask about all the monkey patching in ``testsuite/ext/pylint_compat.py``... :wink:

I will try to cleanup the commits a bit before merging.

AlexanderS:
I think this is ready for merge now.

(With #371 in mind, the tests for the server are excluded for Python2.4 and Python2.5.)

AlexanderS:
Python3 compatibility seems to be a bit more work (see https://travis-ci.org/Bcfg2/bcfg2/jobs/216347528):
  * there a some unicode errors
  * our current pylint/astroid version is not compatible with Python3.5

So I would delay the tests with Python3 for another pull request and merge this as it is.

AlexanderS:
python-augeas fixed the problems with the new version. I think we can merge this now. @solj?

solj:
@AlexanderS Do the python3 tests work already or will that need to be tested later?

AlexanderS:
@solj Python3 would be a different target. We can enable the tests with [allow_failures](https://docs.travis-ci.com/user/customizing-the-build#rows-that-are-allowed-to-fail) so that we can at least see the errors.

AlexanderS:
So we now have working tests again on the current travis infrastructure.

All failures in the Python3 tests are ignored. Most errors are because of the unicode semantic in python3 and some errors seem to be that the Mock module is handled differently.

Additionally we need a newer pylint version with Python3 support that introduces a lot of more warnings/errors. There already was some progress in #269 but it seems to be stalled.

Last but not least we may need to think about alternative packages for missing stuff like m2crypto.

I think we can merge this now and create a new bug/pr for Python3 support.

solj:
@AlexanderS Sounds good to me. Merged. Thanks!

PR #298, opened 2015-07-24
Admin: Add expirecache command, remove it from Info

AlexanderS:
The expirecache command of bcfg2-info was pretty useless, because bcfg2-info
will start a new server process and expire cache cleared the cache in that
process but does not communicate with a running server.

Now expirecache is part of bcfg2-admin and will call expire_metadata_cache
from Core via XML-RPC. This will expire the metadata cache of the supplied
clients, or the complete cache if no arguments are supplied.

### DIFF
--- a/src/lib/Bcfg2/Server/Core.py
+++ b/src/lib/Bcfg2/Server/Core.py
@@ -114,7 +114,8 @@ class DefaultACL(Plugin, ClientACLs):
     def check_acl_ip(self, address, rmi):
         return (("." not in rmi and
                  not rmi.endswith("_debug") and
-                 rmi != 'get_statistics') or
+                 rmi != 'get_statistics' and
+                 rmi != 'expire_metadata_cache') or
                 address[0] == "127.0.0.1")
 
 # in core we frequently want to catch all exceptions, regardless of
--- a/src/lib/Bcfg2/Server/Core.py
+++ b/src/lib/Bcfg2/Server/Core.py
@@ -1366,6 +1366,21 @@ class Core(object):
         return "This method is deprecated and will be removed in a future " + \
             "release\n%s" % self.fam.set_debug(debug)
 
+    @exposed
+    def expire_metadata_cache(self, _, hostnames=None):
+        """ Expire the metadata cache for one or all clients
+
+        :param hostnames: A list of hostnames to expire the metadata
+                          cache for or None. If None the cache of
+                          all clients will be expired.
+        :type hostnames: None or list of strings
+        """
+        if hostnames is not None:
+            for hostname in hostnames:
+                self.metadata_cache.expire(hostname)
+        else:
+            self.metadata_cache.expire()
+
 
 class NetworkCore(Core):
     """ A server core that actually listens on the network, can be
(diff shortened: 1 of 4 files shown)


stpierre:
I think the ``bcfg2-info expirecache`` command was primarily useful in the interactive bcfg2-info interpreter. Can we keep that command and document the difference between the two?

AlexanderS:
But even an active session of bcfg2-info debug would not be influenced by a new bcfg2-info expirecache, or am I missing something?

solj:
@AlexanderS But it would work if you got out of the debug shell and ran the `expirecache` command. I agree that this is useful because it allows the user to see how it will influence a running server core.

stpierre:
Bcfg2-info has its own interactive shell apart from the debug shell.
Expirecache can be used there.

AlexanderS:
Seems like the `bcfg2-info shell` does not work anymore. Only a few commands are documented and no bcfg2-info command seems to work:

```
Welcome to bcfg2-info
Type "help" for more information
bcfg2-info> help

Documented commands (type help <topic>):
========================================
EOF  event_debug  eventdebug  exit  help  quit  update

bcfg2-info> config
*** Unknown syntax: config
bcfg2-info> 
```

Oh and seems like we have to update the bcfg2-info docs http://docs.bcfg2.org/dev/server/bcfg2-info.html because it does not work without an argument.

solj:
Yeah, it does look like bcfg2-info in master needs some work.

AlexanderS:
`bcfg2-info shell` is working again. This now adds the `expirecache` command to bcfg2-admin but also keeps it in the `bcfg2-info shell`. This also fixes a problem with the `Bcfg2.Options.Parser`, that was using a wrong `Namespace` as cache on multiple calls of the same command with the same arguments.

src/lib/Bcfg2/Server/Core.py
```diff
@@ -1366,6 +1366,21 @@ def set_fam_debug(self, address, debug):
         return "This method is deprecated and will be removed in a future " + \
             "release\n%s" % self.fam.set_debug(debug)
 
+    @exposed
+    def expire_metadata_cache(self, _, hostnames=None):
```
stpierre: Access to this function should be denied in the DefaultACL class, since it provides a way for a remote attacker to incur a (possibly significant) performance penalty on the server.

src/lib/Bcfg2/Server/Info.py
```diff
@@ -877,3 +863,39 @@ def run(self):
...
+
+    def do_eventdebug(self, _):
+        """ eventdebug - Enable debugging output for FAM events """
+        self.core.fam.set_debug(True)
+
+    do_event_debug = do_eventdebug
+
+    def do_update(self, _):
+        """ update - Process pending filesystem events """
+        self.core.fam.handle_events_in_interval(0.1)
```
stpierre: Since you've added ``only_interactive``, I suppose most of these (quit, eventdebug, update) could be switched over to command classes in the future if we wanted to.

stpierre:
LGTM once access to expire_metadata_cache is denied by default.

AlexanderS:
expire_metadata_cache should be denied by default.

I converted EventDebug, Update and Quit to command classes and add an alias feature for subcommands.

stpierre:
LGTM.

AlexanderS:
There seems to be an issue with a new version of m2crypto. :(

solj:
Merged, thanks!

PR #299, opened 2015-07-27
Expire metadata cache in all Connector plugins.

AlexanderS:
The plugins need to expire the metadata_cache, even if they only implement `Bcfg2.Server.Plugin.interfaces.Connector.get_additional_data`. This expires the
metadata_cache in all plugins and updates to docs.

The AWSTags plugin is incompatible with aggressive caching mode (like PuppetEnc).
Maybe we should do something like with Packages. The plugins cache the results
indefinitely and it can be refreshed with an xcmd command, but that would be another
pull request.

### DIFF
--- a/doc/development/plugins.txt
+++ b/doc/development/plugins.txt
@@ -118,7 +118,9 @@ Invalidating Caches
 In Bcfg2 1.3.0, some limited :ref:`server-caching` was introduced.  If
 you are writing a :class:`Bcfg2.Server.Plugin.interfaces.Connector`
 plugin that implements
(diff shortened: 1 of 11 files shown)


solj:
Do all of these affect the client Metadata? I'm wondering how useful it will be to expire the metadata cache if, for example, something in SSHbase is updated.

AlexanderS:
Yes, the SSHbase plugin is a connector that adds the public key files to the metadata of a client, so if some file change, we have to update the client metadata. All results of `get_additional_data` is contained in the client metadata and all connector plugins have to expire the cache.

solj:
:+1:

PR #301, opened 2015-07-27
Client/Tools/Systemd: Find extra entries

AlexanderS:
Add FindExtra to get a listing of all extra systemd entries.

### DIFF
--- a/src/lib/Bcfg2/Client/Tools/Systemd.py
+++ b/src/lib/Bcfg2/Client/Tools/Systemd.py
@@ -2,6 +2,8 @@
 
 """This is systemd support."""
 
+import glob
+import os
 import Bcfg2.Client.Tools
 import Bcfg2.Client.XML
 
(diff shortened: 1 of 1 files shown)


src/lib/Bcfg2/Client/Tools/Systemd.py
```diff
@@ -91,3 +93,15 @@ def InstallService(self, entry):
             return self.cmd.run(cmd).success
         else:
             return True
+
+    def FindExtra(self):
+        """Find Extra Systemd Service entries."""
+        specified = [self.get_svc_name(entry)
+                     for entry in self.getSupportedEntries()]
+        extra = set()
+        for fname in glob.glob("/etc/systemd/system/*.wants/*"):
```
solj: Doesn't this directory specified the dependencies of a service? If a service had improperly specified wants, isn't it possible that the entries in the .wants directory may not exist on the system?

AlexanderS: No, the *.wants directories contain symlinks for enabled services. There is no other good way to list all enabled services. For example `systemctl list-unit-files | grep enabled` contains duplicates because it lists all units _and_ all aliases (on my hosts it contain for example ssh and sshd).

solj:
:+1:

stpierre:
@jsbillings @rjsm Can you guys take a quick look at this? Thanks!

rjsm:
At first glance, I think that's an acceptable approach.  I'm curious if systemctl list-units would do what you want, though.

AlexanderS:
@rjsm No `systemctl list-units` does not work, it lists _all_ units that are loaded. That contains a lot units that you do not want to manage with bcfg2. The `systemctl list-unit-files | grep enabled` would be the best way using systemctl, if it would omit the alias names.

PR #302, opened 2015-07-28
BuiltinCore: Fix logging with daemonization

AlexanderS:
python-daemon closes all open files, so we need to specify the filehandles
of the log files to keep logging working.


solj:
:+1:

PR #304, opened 2015-08-07
doc: update Genshi 'name' attribute description

s19n:
Seems current description was added in https://github.com/Bcfg2/bcfg2/commit/255046b3f3484219bbfa6646c88bbc4bbc1b5256 , but is not valid anymore.
Corresponding code:
https://github.com/Bcfg2/bcfg2/blob/master/src/lib/Bcfg2/Server/Core.py#L633
`name` corresponds to `altsrc` only when `realname` attribute is also present; and in https://github.com/Bcfg2/bcfg2/blob/master/src/lib/Bcfg2/Server/Plugin/helpers.py#L107 `realname` has the precedence over `name`.


solj:
:+1:

PR #305, opened 2015-09-09
Keep trailing newlines in Jinja2 templates

gordonmessmer:
Jinja2 will, by default, trim a trailing newline from templates.  This might break some files, such as the iptables state file.

The "keep_trailing_newline" argument can be used to change that behavior.  The drawback is that the option was added in Jinja2 2.7, whereas RHEL 6 includes 2.2.

### DIFF
--- a/doc/releases/1.4.0pre2.txt
+++ b/doc/releases/1.4.0pre2.txt
@@ -19,6 +19,7 @@ environments.
 
 * NagiosGen: Add bundles to configuration
 * HomeBrew: Initial add of plugin
+* Rules/Defaults: Add possibility to use name of entry in attributes
 
 backwards-incompatible user-facing changes
 ------------------------------------------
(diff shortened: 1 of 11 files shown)


stpierre:
If this is only available in newer Jinja2 libraries, I'd rather we only use it if it's supported. E.g., the naive approach would be something like:

    try:
        self.environment = self.__environment_cls__(loader=self.loader,
                                     keep_trailing_newline=True)
    except SomeException:
        self.environment = self.__environment_cls__(loader=self.loader)
        # and probably log a warning

AlexanderS:
Maybe you want to rewrite the history to remove the wrong commit.

gordonmessmer:
Yeah, I was in a rush on my way out of the office.

Try blocks are a little expensive, so I'll look at using distutils to check the Jinja2 version, and set a variable to control the keyword arg use.

solj:
Closing as duplicate of #306

PR #306, opened 2015-09-10
Keep trailing newlines in Jinja2 templates.

gordonmessmer:
Jinja 2 strips a trailing newline from templates by default.  In 2.7, an option was introduced to change this behavior.  Removing the newline breaks some configuration files, such as the iptables state file.


gordonmessmer:
I'm not sure why distutils.version doesn't exist in the test environment.  Should we go with the try/except block?

stpierre:
Yeah, let's do. IMO it's more Pythonic to check behavior anyway, and AIUI this only happens once, at class instantiation time, so the performance hit (if any) of a try-except is minimal.

gordonmessmer:
Reverted the last changes.

solj:
:+1:

PR #307, opened 2015-09-14
Include recent changes in RPM spec dependencies.

gordonmessmer:
These changes are required in order for bcfg2 to run if installed from a clean, minimal installation.

### DIFF
--- a/misc/bcfg2.spec
+++ b/misc/bcfg2.spec
@@ -18,11 +18,11 @@
 # characters from the appropriate line below.
 #
 # Don't forget to change the Release: tag below to something like 0.1
-%global _rc 1
-#%%global _pre 1
+#%%global _rc rc1
+#%%global _pre pre1
 %global _nightly 1
 %global _date %(date +%Y%m%d)
-%global _pre_rc %{?_pre:pre%{_pre}}%{?_rc:rc%{_rc}}
+%global _pre_rc %{?_pre:%{_pre}}%{?_rc:%{_rc}}
 
 # cherrypy 3.3 actually doesn't exist yet, but 3.2 has bugs that
 # prevent it from working:
--- a/misc/bcfg2.spec
+++ b/misc/bcfg2.spec
@@ -18,8 +18,8 @@
 # characters from the appropriate line below.
 #
 # Don't forget to change the Release: tag below to something like 0.1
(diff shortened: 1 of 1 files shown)


misc/bcfg2.spec
```diff
@@ -137,6 +139,8 @@ BuildRequires:    systemd-units
 Requires:         python-ssl
 %endif
 Requires:         libselinux-python
+Requires:         pylibacl
```
stpierre: pylibacl is only used by the client, and it's optional. Bcfg2 should run just fine without it.

gordonmessmer:
The same is true of libselinux-python, so I assumed they'd be treated the same.  Is that the only change you'd like me to back out?

stpierre:
It looks like we're quite inconsistent about what we treat as required/optional in the spec file, and this actually is in the right section, so you can leave it. LGTM as-is.

misc/bcfg2.spec
```diff
@@ -18,8 +18,8 @@
 # characters from the appropriate line below.
 #
 # Don't forget to change the Release: tag below to something like 0.1
-#%%global _rc 1
-%global _pre pre1
+%global _rc 1
+#%%global _pre 1
```
solj: Aren't these set properly by the _pre_rc bits below. It looks to me like that's what the build script uses when creating a new release.

gordonmessmer: They probably should be commented, yes.  

A couple of things look broken, though.  The "release" tag has a macro that requires a number, not "rc" or "pre" suffixed with a number.   If the export script defines _pre as "pre1", the macro will expand that to "prepre1".  

Seems like a bad idea to modify that file in place, too.  Inevitably, someone will run export to tag an rc or pre release, and then later check che changes in to git, unintentionally.  That's probably why _pre is uncommented now.

solj:
It appears as though your master branch also contains some unrelated jinja changes. Perhaps you want to create a couple topical branches for each of these?

gordonmessmer:
I did previously submit a prior pull request from this master branch, but those were merged.  In this pull request, I only see misc/bcfg2.spec being modified.

solj:
My apologies. My local branch wasn't up to date. Merged, thanks!

PR #308, opened 2015-09-14
Prevent lxml Element from being passed non-string data

gordonmessmer:


stpierre:
LGTM.

solj:
Is there a particular plugin that's setting the revision to a non-string character? This seems fine to me but it doesn't make sense to set it to anything other than a string initially afaict.

gordonmessmer:
I suspect it's coming from the Hg plugin, but the core should probably be robust enough to correct simple type errors.

solj:
Merged, thanks!

PR #309, opened 2015-09-17
Reporting: Do not fail, if no interactions found

AlexanderS:
With the filter feature of bcfg2-reports it is possible that all
interactions of a client gets filtered. Now we simply skip those clients
without interactions and do not fail with an exception.


solj:
:+1:

PR #310, opened 2015-09-21
Logging: Fix message building

AlexanderS:
The `logging.LogRecord` instance could contain `args` to replace some
placeholders in the format string in `msg`.


solj:
:+1:

PR #311, opened 2015-09-22

AlexanderS:
This replaces #282.

I squashed the commits by @feuri a bit for cleanup, added the functionality to update from south to django migrations and clean up the django handling.

The upgrade from any 1.3.x version and any future version to the new django migrations should be possible (if the names of the migrations of south and django match).


solj:
:+1:  This all is working beautifully for me now. Nice work!

solj:
@stpierre Does this look good to you? I think this is the main thing missing for us to get a new prerelease out.

AlexanderS:
bcfg2-reports also works now.

reports/reports.wsgi
```diff
@@ -18,4 +16,9 @@ def application(environ, start_response):
         Bcfg2.Options.get_parser().parse()
         config_parsed = True
 
-    return django.core.handlers.wsgi.WSGIHandler()(environ, start_response)
+    if django.VERSION[0] == 1 and django.VERSION[1] >= 7:
+        from django.core.wsgi import get_wsgi_application
+        return get_wsgi_application()(environ, start_response)
+    else:
+        import django.core.handlers.wsgi
+        return django.core.handlers.wsgi.WSGIHandler()(environ, start_response)
```
stpierre: Instead of checking the version, I think this would be more Pythonic to do a try/except -- if that's possible.

src/lib/Bcfg2/Reporting/Storage/DjangoORM.py
```diff
@@ -24,6 +24,16 @@
 from Bcfg2.Reporting.Compat import transaction
 
 
+def get_all_field_names(model):
+    if django.VERSION[0] == 1 and django.VERSION[1] >= 8:
```
stpierre: Is this supposed to reference Django 1.8?

src/lib/Bcfg2/Reporting/templatetags/bcfg2_tags.py
```diff
@@ -394,6 +394,13 @@ def __init__(self, sort_key, text):
         self.sort_key = template.Variable(sort_key)
         self.text = template.Variable(text)
 
+    def _render_template(self, context):
+        if django.VERSION[0] == 1 and django.VERSION[1] >= 8:
+            return context.template.engine.from_string(self.__TMPL__)
+        else:
+            from django.template.loader import get_template_from_string
```
stpierre: Can we move this import outside of the function? Then we can get away from the version comparison, too.

src/lib/Bcfg2/Server/models.py
```diff
@@ -56,15 +57,23 @@ def load_models(plugins=None):
...
+            def __str__(self):
+                return "version %d updated %s" % (self.version,
+                                                  self.updated.isoformat())
+
+            class Meta:  # pylint: disable=C0111,W0232
+                app_label = "reports"
+                get_latest_by = "version"
+        INTERNAL_DATABASE_VERSION = InternalDatabaseVersion
+
+    return INTERNAL_DATABASE_VERSION.objects
```
stpierre: I'm not sure I understand why this was changed. Can you explain a bit?

stpierre:
This looks really, really good. At least, as good as possible given the amount of spaghetti code that has to be introduced. We should seriously look into limiting our Django support in the near future.

I'd like to see fewer explicit version comparisons where possible, although I understand that a lot of them are unavoidable. A few other minor things here and there.

It'd also be nice to squash some of our commits to have a nice history, if possible.

src/lib/Bcfg2/Reporting/Storage/DjangoORM.py
AlexanderS: Yes: https://docs.djangoproject.com/en/1.8/releases/1.8/#model-meta-api

I know that "django 1.7" is in the title of the pull request, but thanks to @solj we also added support for 1.8.

src/lib/Bcfg2/Reporting/templatetags/bcfg2_tags.py
AlexanderS: The import fails on current django versions and `context.template.engine` does not exist in older versions. We could use try/except, but I do not thing that will be better.

src/lib/Bcfg2/Server/models.py
AlexanderS: I do not know how this was supposed to work. Classes defined in functions are not available at global scope: 

```
>>> def foo():
...     class Bar():
...         pass
... 
>>> foo() 
>>> Bar()
  File "<stdin>", line 1, in <module>
NameError: name 'Bar' is not defined
```

AlexanderS:
Squashing is no problem, but I will wait until we are ready to merge. It is easier to review only new commits.

reports/reports.wsgi
AlexanderS: Fixed.

solj:
This looks good. I also agree with @stpierre that we should limit our Django support. I will need to look at which distros support which versions. I no longer see django available in EPEL for centos 5 which was the reason for supporting Django 1.4. Perhaps we should increase that? But to what?

stpierre:
I could pick a few more nits, but overall this is a very high quality patch. Let's get it merged.

THANK YOU for your tireless work and longsuffering!

solj:
Merged, thanks!

solj:
Whoops forgot to wait for the squash. My fault. I really want to get the 1.4 prerelease out though.

PR #317, opened 2015-09-29
Run the superclass shutdown() to correctly set self.started

kincl:
The superclass sets self.started to False which makes sure the logic works correctly and notifier.stop() does not get run more than once.


stpierre:
LGTM

solj:
Merged, thanks!

PR #318, opened 2015-09-30
Bundler: Fix addition of bundle.

AlexanderS:
`bundles` is a set and does only have `add()` but not `append()`.


solj:
Merged, thanks!

PR #320, opened 2015-11-17
Some more django fixes

AlexanderS:
Here are some more django fixes:

* Fix the database backend for the Metadata plugin (requires the same handling for django1.7 like the reporting database backend with `django.setup()`)
* Add a remaining fix for django1.5, that was missing in 84a83ca. (The first parameter of the `url` template tag has to be quoted.)
* Use consistent syntax for the `url` template tag and remove deprecated (in django1.8) syntax. The `url` template tag should be called with an identifier and not with the python dotted path of a view function.

Hopefully the last django1.7 fixes ;)


solj:
:+1:

PR #321, opened 2015-11-17
Client/Tools: Fix error if no pkg handled

AlexanderS:
`Install()` of a client tool returns the changed states as dict, that will update
the global state. If nothing is handled withing a tool, it have to return an empty
dict and not None because the return value is directly used as argument for update:

> APT.Install() call failed:
> Traceback (most recent call last):
>   File "/usr/lib/python2.7/dist-packages/Bcfg2/Client/__init__.py", line 739, in DispatchInstallCalls
>    self.states.update(tool.Install(handled))
> TypeError: 'NoneType' object is not iterable


solj:
Merged, thanks!

PR #326, opened 2016-01-20
testsuite: Require newer version of mock and change import

AlexanderS:
mock-1.1.0 was converted to a package and needs to be imported differently.


solj:
:+1:

stpierre:
Looks like mock's old CamelCase assertions have been removed, and we've got at least one of those still in the code, so that'll need to be fixed as well.

No RPM-based distribution has a new enough version of mock, so we'll need to comment out the %check section of the spec files.

AlexanderS:
Oh maybe it's better to try/catch the two imports and do not require the newer version than. I will fix the CamelCase issue.

solj:
Ah, even better. Now there's no need to comment out the RPM checks.

solj:
Merged, thanks!

PR #327, opened 2016-02-17
Setup options

gordonmessmer:
After getting another traceback due to a reference to a non-existent option, I crawled the entire code base to look for options that I couldn't find at least one place where they were set.

I'm not at all certain of these fixes because the code doesn't seem to match the documentation, and I could be wrong about any part of this.  Most of this is based on my understanding of the code, and not on testing.  All the same...

The ldap module refers to ldap_retries and ldap_retry_delay, but as far as I can tell, those aren't the names that the options are set to.  I added dest= to fix that.

Similarly, cfg_validation doesn't look like it uses that name unless dest= is set.

In MultiprocessingCore.py and Core.py, a "location" option is referenced, but the same option seems to be referenced as "server" elsewhere, and I'm pretty sure that's the correct name.

A few files refer to "servicemode", but the correct option is "service_mode".

Finally there are a couple of settings that I can't figure out at all.

setup.decision_list is referenced in two files, but I can't find it set anywhere:
src/lib/Bcfg2/Client/Tools/__init__.py:                    entry, Bcfg2.Options.setup.decision_list)):
src/lib/Bcfg2/Client/Tools/__init__.py:                    entry, Bcfg2.Options.setup.decision_list)):
src/lib/Bcfg2/Client/__init__.py:                    Bcfg2.Options.setup.decision_list = \
src/lib/Bcfg2/Client/__init__.py:                    self.logger.info(Bcfg2.Options.setup.decision_list)
src/lib/Bcfg2/Client/__init__.py:                dwl = Bcfg2.Options.setup.decision_list
src/lib/Bcfg2/Client/__init__.py:                     passes_black_list(e, Bcfg2.Options.setup.decision_list)]

setup.hm is referenced in tools/git_commit.py.  I can't find that setting anywhere, and git_commit.py appears to reference other non-existent options, as well, such as profile=Bcfg2.Options.CLIENT_PROFILE.

src/lib/Bcfg2/Client/Proxy.py references setup.ssl_cns, but it looks like that option would be setup.serverCommonNames.

...And in the course of reviewing the code, I found what seems like a *lot* of options that are parsed, but never used, which I'm sure is terribly misleading.

I'd really like to see someone who understands option handling better than I do make a pass over these things and make the necessary changes.

Systemd: Unexpected failure installing Service:tftp.socket
  File "/usr/lib/python2.7/site-packages/Bcfg2/Client/Tools/__init__.py", line 242, in Install
    states[entry] = func(entry)
  File "/usr/lib/python2.7/site-packages/Bcfg2/Client/Tools/Systemd.py", line 79, in InstallService
    if Bcfg2.Options.setup.servicemode == 'disabled':
AttributeError: 'Namespace' object has no attribute 'servicemode'


solj:
It looks to me like decision_list is set at https://github.com/Bcfg2/bcfg2/blob/master/src/lib/Bcfg2/Client/__init__.py#L319-L322

solj:
Looks like the tools/git_commit.py tool is using the old options parsing code. That will need to be migrated to the new options parsing.

gordonmessmer:
re: decision_list.  Yeah, I don't know why I just wasn't seeing it.  It's right there in the "grep" output I included.  Toward the end of the exercise I was feeling pretty frustrated, especially with the number of (apparently?) unused Options.

solj:
All the changes here look good to me. I still need to investigate the ssl_cns option from Proxy.py.

AlexanderS:
> src/lib/Bcfg2/Client/Proxy.py references setup.ssl_cns, but it looks like that option would be setup.serverCommonNames.

That looks correct.

solj:
@AlexanderS Okay. In that case, can you add that change @gordonmessmer ?

gordonmessmer:
I set dest for ssl_cns.

gordonmessmer:
Assuming this gets merged, I'll try to find time to crawl the code again and remove unused options for a later pull request.

solj:
Merged, thanks!

PR #328, opened 2016-02-22
Options: migrated tools/git_commit.py to new parser

AlexanderS:
As noticed in #327 ``tools/git_commit.py`` should be converted to the new Options parser. This should do it. It also changes the config format a bit, because the logging stuff is now in ``[logging]`` and not in ``[global]`` anymore.


solj:
:+1:

AlexanderS:
One thing: I do not know if it is possible to overwrite the logging options from ``bcfg2.conf`` this way. I do not know if the original config (specified in ``[global]`` ``config``) is loaded before evaluating the other config options or if it is loaded later and overwrites the settings specified in in this configuration.

PR #329, opened 2016-03-31
Allow multiline scripts with indentation in bcfg2-info debug mode

s19n:
This should allow to execute python scripts with indentation through bcfg2-info debug command.

PR #330, opened 2016-03-31
testsuite: Change imports for newer mock versions (maint)

AlexanderS:
This is a backport of #326 for maint. So that we can fix the test cases for other fixes like #329.


AlexanderS:
oO There are two new failing test cases.

AlexanderS:
Found that strange bug. (Could not replicate it on my local machine.) Apparently some python versions pass `self` even if you call copies of unrelated methods from member attributes. This is fixed by wrapping the copied function in `staticmethod`.

PR #333, opened 2016-04-21
Fix a variety of problems with YumHelper

gordonmessmer:
With the new options code, classes used as the parent argument to register_commands must not access the setup object in their init function, because that object hasn't been prepared yet. It's best, then, not to have an init function at all.

Various other changes were made to get bcfg2-yum-helper working again.

### DIFF
--- a/src/lib/Bcfg2/Server/Plugins/Packages/Yum.py
+++ b/src/lib/Bcfg2/Server/Plugins/Packages/Yum.py
@@ -456,16 +456,13 @@ class YumCollection(Collection):
                 reponame = basereponame
 
                 added = False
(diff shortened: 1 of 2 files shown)


gordonmessmer:
Ping

solj:
I haven't used the YumHelper but this looks fine to me :+1:

gordonmessmer:
Anyone else?

solj:
@AlexanderS I assume you're not using the YumHelper either but does this look good to you?

AlexanderS:
Yes, looks fine to me :+1:

solj:
Merged, thanks!

PR #334, opened 2016-06-29
Don't use 'input' function under python2.

gordonmessmer:
If diff output contains non-ascii characters, then the first call to safe_input in prompt() will raise UnicodeEncodeError.  prompt will then use input(), which will fail under python2.

This change uses safe_input() in both cases.


AlexanderS:
:+1:

solj:
Merged, thanks!

PR #335, opened 2016-07-12
Some small fixes for the docs

AlexanderS:
This are only some small fixes for the docs. Some small formating issues and an update for the lax_decryption setting (The config setting was changed in 919059f0971c0f8bf18ca2cedb3c7e5319632726).


solj:
:+1:

PR #336, opened 2016-07-15
Ldap enhancements

AlexanderS:
This adds a few more things to the LDAP plugin and removes the issues mentioned in the docs:
- It supports ldaps now, by providing the possibility to set the LDAP URI directly and setting arbitrary options. 
- The plugin does work with all metadata cache modes, because it only caches the available queries in the metadata cache and adds a separate cache for the results. By default it will cache the results until the next client run. But it can be configured by config file, that it should not cache anything (the ldap query is executed on every access) or that it should keep the cache until it is expired manually via XMLRPC. The cache can be expired by query, by hostname or completely.

### DIFF
--- a/doc/server/plugins/grouping/ldap.txt
+++ b/doc/server/plugins/grouping/ldap.txt
@@ -87,6 +87,26 @@ If you wish, you could customize these values in your ``bcfg2.conf``::
     retries = 3
     retry_delay = 3.0
 
+Caching
++++++++
+
+This module could not know, if a value changed on the LDAP server. So it does not cache
+the results of the LDAP queries by default.
+
+You could enable the cache of the results in your ``bcfg2.conf``:
+
+    [ldap]
+    cache = on
+
+If you enable the caching, you have to expire it manually. This module provides a XML-RPC
+method for this purpose: :func:`Ldap.expire_cache
+<Bcfg2.Server.Plugins.Ldap.expire_cache>`.
+
+Even without enabling caching, the results of the LDAP queries are cached, but are
+discarded before each client run. If you access the Ldap results of different client, you
+may get cached results of the last run of this client. If you do not want this behaviour,
+you can disable the caching completely by setting it to ``off``.
+
 Class reference
 ---------------
 
@@ -251,9 +271,3 @@ Known Issues
 ------------
 
 * At this point there is no support for SSL/TLS.
-* This module could not know, if a value changed on the LDAP server. So it could not
-  expire the client metadata cache sanely.
-  If you are using aggressive caching mode, this plugin will expire the metadata cache
-  for a single client at the start of a client run. If you are using LDAP data from
-  another client in a template, you will probably get the cached values from the last
-  client run of that other client.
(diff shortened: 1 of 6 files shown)


doc/server/plugins/grouping/ldap.txt
```diff
@@ -87,6 +87,26 @@ If you wish, you could customize these values in your ``bcfg2.conf``::
...
+
+This module could not know, if a value changed on the LDAP server. So it does not cache
+the results of the LDAP queries by default.
+
+You could enable the cache of the results in your ``bcfg2.conf``:
+
+    [ldap]
+    cache = on
+
+If you enable the caching, you have to expire it manually. This module provieds a XML-RPC
```
solj: Just a small typo for "provides"

src/lib/Bcfg2/Server/Plugin/helpers.py
```diff
@@ -1698,3 +1698,84 @@ def AddDirectoryMonitor(self, relative):
...
+
+
+class CallableDict(MutableMapping):
+    """ This maps a set of keys to a set of value-getting functions;
+    the values are populated on-the-fly by the functions as the values
+    are needed (and not before).  This is for example used by
+    :func:`Bcfg2.Server.Plugins.Packages.Packages.get_additional_data`;
+    see the docstring for that function for details on why.
+
+    Unlike a dict, you can specify values or functions for for the
```
solj: Typo "for for"

solj:
This looks good to me :+1:

AlexanderS:
Fixed the two typos.

src/lib/Bcfg2/Compat.py
```diff
@@ -292,3 +292,19 @@ def cmp(a, b):
     from ast import literal_eval
 except ImportError:
     literal_eval = eval
+
```
solj: @AlexanderS Do we need to add this? We are requiring 2.6 or later for the server-side, correct?

AlexanderS: I think we never officially deprecated python <= 2.5 on the server side (at least according to the docs: https://github.com/Bcfg2/bcfg2/blob/master/doc/installation/prerequisites.txt#L51).

I thought the plan was to drop support for older python versions on the server with bcfg2-1.4. If we decide to drop it, we can maybe de-compat some more stuff on the server side.

solj: Interesting. In any case, those prerequisites are wrong. There's no way 2.2 will ever work with the current code. I believe we should require 2.6 or later on the server since it should be easy enough for folks to run a server on a recent distro (even if they do have to support some older clients still).

AlexanderS:
Removed the `partial` compat.

solj:
@AlexanderS This looks good to me. Do you have any additional items to add to this? Also, should we perhaps add this to the release notes so we can go ahead and merge?

AlexanderS:
I think, we can merge this as it is and I will create a new PR with 1.4.0rc1 release notes (containing all the changes since 1.3 and a short section about differences from 1.4.0pre2).

solj:
Merged, thanks!

PR #337, opened 2016-07-17
Helper scripts packed in bcfg2-utils

AlexanderS:
This is a replacement for #132 on top of master, with the additional changes suggested in the old pull request and a small change to include the bash_completion in the bcfg2-server package.


solj:
Looks good, thanks!

PR #339, opened 2016-08-01
doc: Clarification on negated groups

s19n:
From IRC: 
> AlexS[zedat]: This is a known limitation. Bundles can not be inside conditionals in Metadata/groups.xml.


solj:
Merged, thanks!

PR #340, opened 2016-08-11
Add support for rawurl in Apt

rbuch:


src/lib/Bcfg2/Server/Plugins/Packages/Apt.py
```diff
@@ -34,8 +35,12 @@ def get_config(self):
 
         for source in self:
             if source.rawurl:
-                self.logger.info("Packages: Skipping rawurl %s" %
-                                 source.rawurl)
+                if source.rawurl[-1] != '/':
+                    source.rawurl = source.rawurl + "/"
+                index = string.rfind(source.rawurl, "/", 0, -1)
```
AlexanderS: Maybe you mean:

```
index = source.rawurl.rfind('/', 0, -1)
```

`string` is meant to by the variable holding the string.

rbuch: It should work either way:

```
>>> import string
>>> string.rfind("testtest", "s", 0, 8)
6
>>> "testtest".rfind("s", 0, 8)
6
```

I can change it if you think it's better style.

rbuch:
I'm not sure why the Travis CI build is failing now, the 2.7 builds are passing, but the 2.6 builds seem to have malformed `pip` commands.

src/lib/Bcfg2/Server/Plugins/Packages/Apt.py
AlexanderS: Yes, I would change that. The [python docs](https://docs.python.org/2/library/string.html#deprecated-string-functions) states, that you should consider this functions in the string library as deprecated.

AlexanderS:
Yes seems like travis updated the images and the new pip refuses `--use-mirrors`. #297 would fix this, but I would like to cleanup the commits in #297 before merging.

src/lib/Bcfg2/Server/Plugins/Packages/Apt.py
rbuch: I've fixed this in the latest commit.

AlexanderS:
You can rebase this now on master to fix the build.

solj:
@AlexanderS Looks like the checks pass now. :+1: for me.

PR #341, opened 2016-08-17
travis-ci: Fix installation of dependencies

AlexanderS:
This is just a quick fix: The new version of pip on travis does
fail when using the `--use-mirrors` option (to be fair, the option
was deprecated long time ago).

The real fix will be #297 (once it will be ready to merge).

### DIFF
--- a/testsuite/install.sh
+++ b/testsuite/install.sh
@@ -10,13 +10,13 @@ pip install -r testsuite/requirements.txt
 PYVER=$(python -c 'import sys;print(".".join(str(v) for v in sys.version_info[0:2]))')
 
 if [[ ${PYVER:0:1} == "2" && $PYVER != "2.7" ]]; then
-    pip install --use-mirrors unittest2
+    pip install unittest2
 fi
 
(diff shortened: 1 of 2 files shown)


testsuite/install.sh
```diff
@@ -10,13 +10,13 @@ pip install -r testsuite/requirements.txt
...
 
 if [[ ${PYVER:0:1} == "2" && $PYVER != "2.7" ]]; then
-    pip install --use-mirrors unittest2
+    pip install unittest2
 fi
 
 if [[ "$WITH_OPTIONAL_DEPS" == "yes" ]]; then
     sudo apt-get install -y yum libaugeas0 augeas-lenses libacl1-dev libssl-dev
 
-    pip install --use-mirrors PyYAML pyinotify boto pylibacl 'django<1.5' \
```
solj: @AlexanderS Should we change the django requirement? We support versions more recent than 1.4.

AlexanderS: Oh yes. Let me fix this.

PR #342, opened 2016-08-17
travis-ci: Use django1.6 for python2.6.

AlexanderS:
Python2.6 support was dropped in django-1.7. This is an addition for #341.


AlexanderS:
Let's wait until travis is ready and the build is green...

AlexanderS:
:tada: And I thought that would be just a simple quick fix...

PR #343, opened 2016-08-17
Fix syntax errors in documentation

AlexanderS:
This fixes various small syntax errors in the documentation. It also adds a unit test to check for syntax errors and warnings (only enabled for the tests with python2.7 and optional dependencies).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,14 @@
 language: python
-python:
-  - "2.6"
-  - "2.7_with_system_site_packages"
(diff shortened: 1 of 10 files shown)


AlexanderS:
This also adds some optional dependencies, that were missing before. Now the python2.7 tests with options dependencies run _all_ tests (except one test, that is unconditionally skipped: `"Tests that call os.fork are broken, even when os.fork is mocked"`).

PR #345, opened 2016-08-18
Fixes for django 1.10

AlexanderS:
Django 1.10 removes the patterns method for building the urlpatterns. This adds a simple fallback method into `Reporting/Compat.py`.

### DIFF
--- a/src/lib/Bcfg2/Reporting/Compat.py
+++ b/src/lib/Bcfg2/Reporting/Compat.py
@@ -13,4 +13,16 @@ try:
     from django.conf.urls.defaults import url, patterns
 except ImportError:
     # Django > 1.6
-    from django.conf.urls import url, patterns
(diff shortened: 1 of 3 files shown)


solj:
I'm wondering if instead of this, we should change urls.py to the new way of doing things. As far as I can tell from reading the documentation, there should be no backwards compatibility issues when not using patterns (unless I'm missing something). https://docs.djangoproject.com/en/1.10/releases/1.8/#django-conf-urls-patterns

AlexanderS:
We can also do that. In the docs for django 1.7 they state:

> Always use patterns() to create the urlpatterns variable.
https://docs.djangoproject.com/en/1.7/topics/http/urls/#syntax-of-the-urlpatterns-variable

But the "new" way should be compatible.

PR #346, opened 2016-08-26
debian/control: Replace python-support with dh-python

0:
According to the [Debian Python FAQ](https://wiki.debian.org/Python/FAQ), python-support has been deprecated in favor of dh-python.


solj:
:+1:

AlexanderS:
Merged, thanks!

PR #350, opened 2016-08-31
Fix the url tags for django 1.9

### LINKED ISSUE #349: python2-django 1.9 error (1.7 works)
In Arch, "python2-django 1.7.11" works fine, but when updating to "python2-django 1.9.8" I get the following which causes an Apache 500 error:
https://bpaste.net/show/b4f0be565a3a
Switching back to 1.7 fixes the issue.

AlexanderS:
This fixes the url template tags for django 1.9 and above. The `future` template tag library was removed and you should use the default url tag now. This should fix #349.

This also adjusts the django version that is required for bcfg2-web. As the future tag lib was introduced with django1.3 this was required before and that change only fixes the specification.

I do not know if we want to support django<1.3 on the server side (what version is available with rhel6?). When dropping django<1.3 on the server side too, we can remove one version switch in `DBSettings.py`.


solj:
@AlexanderS I'm pretty sure nothing older than 1.4 is included in any supported releases so I'd be okay with dropping support for <= 1.3.

PR #352, opened 2016-09-07
Packages: Strip ':any' suffix from APT dependencies

0:
It's possible for dependency names in Debian and Ubuntu packages to end in ':any', which signifies that the architecture doesn't matter. This may cause the following message (or something similar) to be logged:
```
Packages: Got 4 unknown entries
Packages: ['python2.7:any', 'python3:any', 'python:any', 'python3.5:any']
```

This patch just strips the suffix from all dependency names that have it.

### DIFF
--- a/src/lib/Bcfg2/Server/Plugins/Packages/Apt.py
+++ b/src/lib/Bcfg2/Server/Plugins/Packages/Apt.py
@@ -6,6 +6,15 @@ from Bcfg2.Server.Plugins.Packages.Collection import Collection
 from Bcfg2.Server.Plugins.Packages.Source import Source
 
 
+def strip_suffix(pkgname):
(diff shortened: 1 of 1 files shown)


src/lib/Bcfg2/Server/Plugins/Packages/Apt.py
```diff
@@ -59,6 +59,15 @@ class AptSource(Source):
     #: AptSource sets the ``type`` on Package entries to "deb"
     ptype = 'deb'
 
+    @staticmethod
```
AlexanderS: I wouldn't use staticmethod here and simply add a method outside of the class scope on the top level in this file.

0:
Moved the method to the top level.

AlexanderS:
:+1:

solj:
Merged, thanks!

Issue #348, opened 2016-08-31
Package groups are not supported by PacCollection


jsteel44:
For example on Arch, I would like to specify the "base" package group.

jsteel44:
I've tried #354 but see the error "Packages: 'base' is not a valid group"

jsteel44:
Looks good after clearing the cache, thanks!

Issue #347, opened 2016-08-31
Pac-Packages backend does not handle dependencies


jsteel44:
Therefore dependencies are classed as "extra entries".

jsteel44:
I've tried #354 (applied to both client and server) but still seeing dependencies under "extra entries". ie I've got bcfg2 to install "vim", yet "vim-runtime" is an extra entry.

$ pacman -Qi vim-runtime
Name            : vim-runtime
Required By     : vim
Install Reason  : Installed as a dependency for another package
[...]

AlexanderS:
Did you regenerate the package cache? Either by removing the cache files or with `bcfg2-admin xcmd Packages.Refresh`?

jsteel44:
Looks good after clearing the cache, thanks!

PR #354, opened 2016-09-11
Improve Pacman support

0:
* Add dependency resolution (#347).
* Add groups (#348).

I don't know the appropriate way to invalidate the cache here, so it's currently up to the user to remove the old cache file if things break.

### DIFF
--- a/src/lib/Bcfg2/Server/Plugins/Packages/Pac.py
+++ b/src/lib/Bcfg2/Server/Plugins/Packages/Pac.py
@@ -2,6 +2,7 @@
 
 import os
 import tarfile
+from Bcfg2.Compat import cPickle
 from Bcfg2.Server.Plugins.Packages.Collection import Collection
 from Bcfg2.Server.Plugins.Packages.Source import Source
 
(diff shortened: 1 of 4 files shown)


AlexanderS:
:+1:

src/lib/Bcfg2/Server/Plugins/Packages/Pac.py
```diff
@@ -24,13 +76,36 @@ def __init__(self, metadata, sources, cachepath, basepath, debug=False):
...
+
 
 class PacSource(Source):
     """ Handle Pacman sources """
 
     #: PacSource sets the ``type`` on Package entries to "pacman"
     ptype = 'pacman'
 
+    def __init__(self, basepath, xsource):
+        self.pacgroups = dict()
```
solj: While it's commonly used elsewhere in the code, I think it's probably best to use braces ({}) instead of dict(). I think that the performance difference is reason enough and I personally find them equally readable.

Sorry to be pedantic, I just want to choose the best option moving forward.

0: No worries, I'll change it. I also prefer `{}`, but I wanted to be consistent.

There are also some places where `defaultdict` feels appropriate, but it wasn't added until Python 2.5. I didn't see it used anywhere and I wasn't sure what the minimum supported version is on the server, so I avoided it.

0:
Replaced `dict()` with `{}`.

solj:
Merged, thanks!

PR #360, opened 2016-09-24
bcfg2-admin: pass SSL protocol option for perf and xcmd connections

arachnopavel:
bcfg2-admin xcmd and perf are broken after commit a2b8b3282bc07e1db362d2edd51d2bee3e425d57
Ohai probe is unable to run on Ubuntu/Debian systems with /bin/dash as /bin/sh, due to syntax incompatibility with bash/csh


solj:
:+1:

AlexanderS:
:+1:

solj:
Merged, thanks!

PR #362, opened 2016-09-28
Options: Missing config file is not a fatal error

AlexanderS:
If the config file is not available, it should not be a fatal error,
but bcfg2 should just use the default config values.


AlexanderS:
Ok, this failing test should be removed.

solj:
Looks like it's still failing.

AlexanderS:
Ah, this was another problem. The options parser cannot parse the arguments of the testrunner. So we have to explicitly supply custom arguments.

PR #364, opened 2016-10-02
doc: Document new behavior from 5a0307fb

solj:
We no longer strip() the trailing whitespace from Probe output which can
result in different behavior than previous versions.

### DIFF
--- a/doc/releases/1.4.0pre1.txt
+++ b/doc/releases/1.4.0pre1.txt
@@ -43,6 +43,9 @@ backwards-incompatible user-facing changes
     You can convert your existing decisions using
     ``tools/upgrade/1.4/migrate_decisions.py``.
 
+* The Probe plugin no longer strips probe output. You may need to adjust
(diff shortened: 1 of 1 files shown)


AlexanderS:
Is this really an intended change?

Btw: And I cannot find a `strip()` in the old code. Where was it removed? The only change I found in 5a0307f is in the group handling, that is now using a regexp to match the group names and strip trailing whitespaces in one step.

AlexanderS:
Do we really want to document this as change? Or do we want to add a `strip()` so that the output is like before?

solj:
@AlexanderS I believe it was an intended change so that the probes were technically accurate. In other words, I'm guessing someone had a Probe where they *wanted* the whitespace to stick around. I prefer the old way but I can see where one might run into problems if they care about the trailing whitespace.

AlexanderS:
Maybe we want to strip the output by default, but add the possibility to flag the Probes that shouldn't be striped somehow?

PR #367, opened 2016-11-17
Packages: Cleanup pkgnames

AlexanderS:
The pkgnames should be reset before parsing the source files. If a package was
removed for a source, it should be removed from the pkgnames set, too. The
packages plugin should look for the next source containing the package.


solj:
:+1:

PR #368, opened 2016-11-18
Cfg: Fix SSLCA key generation

solj:
The previous check resulted in a "FutureWarning" and also did not work
as expected and reset the key values in every case.


AlexanderS:
:+1:

PR #369, opened 2016-11-21
Lint: Some fixes of the required attributes

### LINKED ISSUE #365: Unknown Path type augeas
If I run `bcfg2-lint` with an augeas Path present, I get the following:
```
ERROR: Unknown Path type augeas:    line 22: <BoundPath type="augeas" .../>
```

I believe the culprit is [the Path dict in RequiredAttrs](https://github.com/Bcfg2/bcfg2/blob/a472e687840a12a20bb2c22875de6efad8ddc257/src/lib/Bcfg2/Server/Lint/RequiredAttrs.py#L56). My current workaround is to add just
```
augeas=dict()
```
because I don't know which attributes are required.

AlexanderS:
This should fix #365. Additionally I tried to update the required attributes of the other path entries:

   * `name` should be required by all path entries (only missing in vcs before)
   * The permission attributes `mode`, `group`, `owner` are required by some more types. (at least it is required by the `__req__` entries in the client tools...)


src/lib/Bcfg2/Server/Lint/RequiredAttrs.py
```diff
@@ -47,32 +47,40 @@ def is_device_mode(val):
...
-                ignore=dict(name=is_filename),
-                nonexistent=dict(name=is_filename),
-                permissions=dict(name=is_filename, owner=is_username,
-                                 group=is_username, mode=is_octal_mode),
-                vcs=dict(vcstype=lambda v: (v != 'Path' and
-                                            hasattr(Bcfg2.Client.Tools.VCS.VCS,
-                                                    "Install%s" % v)),
-                         revision=None, sourceurl=None)),
+            Path={
+                '__any__': dict(name=is_filename),
```
AlexanderS: Done.

solj:
:+1:

PR #371, opened 2016-11-23
Drop support for Python < 2.6 on the server

AlexanderS:
This drops the support for Python < 2.6 on the server. I removed some compat stuff and used the native variants on the server for compat stuff, that is still required for the client code.

Most stuff in `Bcfg2.Compat` is for Python 3 compatibility now. We could consider to use https://pypi.python.org/pypi/six and drop our own Python 3 compat stuff. I think it provides all the Python 3 compatibility we need.


doc/releases/1.4.0pre2.txt
```diff
@@ -46,6 +46,8 @@ backwards-incompatible user-facing changes
   your queries default to the class names and the Ldap plugin expires
   the metadata caches if the config file changes.
 
+* Droped support for Python 2.5 or below on the bcfg2 server.
+
```
solj: Perhaps we should create release notes for pre3 (or rc1, not sure) and add this there.

AlexanderS:
Just a quick note: Because we have now at least 2.6 on the server we could use something like that:
```
from __future__ import print_function, unicode_literals
```
to deal with the Python3 incompatibility.

doc/releases/1.4.0pre2.txt
AlexanderS: Should rc1 release note contain all changes since 1.3.x?

solj: @AlexanderS Hmm, that's a good question. I think you may be right that rc1 should contain all changes. Especially since we haven't released something in a while.

solj:
@AlexanderS I think we just need to exclude the python2.4 travis tests to get the check to succeed here.

AlexanderS:
I fixed the testsuite. The `sbin/` checks had wrong blacklist handling.

PR #372, opened 2016-11-23
Some small fixes for NagiosGen

### LINKED ISSUE #161: NagiosGen: doesn't include default-group.cfg
When I generate nagiosgen.cfg it doesn't include the default-group.cfg file described in the documentation.

AlexanderS:
Some small fixes for the NagiosGen plugin:

* Update documentation: The _backwards-compatibility layer_ was removed.
* Always include default group. This should fix #161.
* Small code cleanup. Why should you use a list if you want to have a set?!


solj:
:+1:

PR #374, opened 2016-12-14
fix the case when user must not have any supplementary groups

arachnopavel:
Fix the case when user must not have any supplementary groups, by supplying the -G option with empty group list.


AlexanderS:
Seems like the test needs to be adjusted.

src/lib/Bcfg2/Client/Tools/POSIXUsers.py
```diff
@@ -261,6 +261,8 @@ def _get_cmd(self, action, entry):
                           for e in entry.findall("MemberOf")]
                 if extras:
                     cmd.extend(['-G', ",".join(extras)])
+                else:
+                    cmd.extend(['-G', ''])
```
solj: Could we achieve this same effect by doing something like https://gist.github.com/solj/a1d89b429e895471f275c1387f67418e? This seems like a simpler route.

arachnopavel: I'd prefer my way because it's more explicit, just in case.

arachnopavel: Is it going to be accepted, or should I refactor this particular change first?

arachnopavel:
Are there any issues with these commits that need to be resolved on my part?

AlexanderS:
FYI: I just cherry picked the XMLSyntaxError commit to master c5c90ee87a93b778117d24bf822152eb7582dca3

solj:
:+1:

AlexanderS:
Merged, thanks!

PR #379, opened 2017-03-07
testsuite: Use python-augeas from PyPI with pip

AlexanderS:
fedorahosted.org was retired on March 1st, 2017. So we need to pull
python-augeas from anywhere else, let's simply install it with pip
from PyPI.


AlexanderS:
This will be required on master, too.

PR #381, opened 2017-03-09
README enhancements

AlexanderS:
If we add the file extension to our README file, GitHub will parse the markup and pretty print it. This will get us clickable links and nice headlines.

With this formatted README we could also add a travis-ci badge.

You can see the rendered README in [my branch](https://github.com/AlexanderS/bcfg2/tree/feature/README).


solj:
:+1:

PR #382, opened 2017-03-09
Fix multiline scripts in bcfg2-info debug mode

AlexanderS:
This should allow to execute python scripts with indentation through bcfg2-info debug command.

(This is #329 rebased on master.)


solj:
:+1:

PR #383, opened 2017-03-17
Server/Lint: Include RequiredBundle in referenced bundles

### LINKED ISSUE #378: Suppress lint warning for unreferenced RequiredBundle
I have a bundle that's only used as a RequiredBundle. I think this lint warning should not happen:

> WARNING: Bundle ... defined, but is not referenced in Metadata

0:
Fixes #378.


AlexanderS:
:+1:

solj:
Merged, thanks!

PR #384, opened 2017-03-21
Server/Info: Fix encoding declaration

AlexanderS:
It seems to be, that there must not be any Python statement before the
magic comment.


solj:
:+1:

PR #385, opened 2017-03-22
Probes: Remove wrong line

AlexanderS:
This line brakes django detection for the Probes! It resets the global
variable just after the detection. This line was left while changing
the placement of the detection in fa0d86aba32c40d829f9f94411403221a48283e8.


solj:
:+1:

PR #388, opened 2017-05-05
Reporting/Storage: Import ServiceEntry

0:
It wasn't included in the transition from a wildcard import in 255faaf.


AlexanderS:
Merged, thanks!

PR #389, opened 2017-05-08
Reporting/Storage: Import ActionEntry

0:
Sorry, I missed this in #388. Are there any others that should be imported here?


AlexanderS:
The problem is, that we currently do not have any python code checks for that part of the source. I started work to enable the check on the django/reporting stuff, but it is not ready yet. I will try to get it done this week, so that we can ensure we import all required models.

AlexanderS:
Merged, thanks!

PR #391, opened 2017-07-14
Add rudimentary windows support for bcfg2

chschenk:
This pull request adds rudimentary windows support to bcfg2. The most changes are in the probe and the locking sections:

-Probes
        The shebangline of the probes have to be in the format "#!C:\Path\to\interpreter.exe*.fileending" because windows refuses to run files without an fileending. An example shebangline would be "#!C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe*.ps1"

-Locking
        Unfortunately windows has no fcntl so that locking under windows is done by writing the pid into the lockfile

-Tools
        For managing windows i have written three Tools: WinAction, WinFS, WinService
        -WinAction handles the default Action entrys. Example entry <Action name='test' command="echo test > C:\test.txt" timing="post" when="always" status="check"/>
        -WinFS handles Path entrys. Example entrys "<Path name='/%ProgramFiles%/test/test.txt'/>" or "<Path name='/C:/test.txt'/>". Currently the options owner, group and mode are ignored completly.
        -WinService handles service entrys and starts, stops or restarts the services via powershell. Example Entry: "<Service name='icinga2' type='windows' status="on" />"

### DIFF
--- a/src/lib/Bcfg2/Client/Tools/WinAction.py
+++ b/src/lib/Bcfg2/Client/Tools/WinAction.py
@@ -1,0 +1,79 @@
+"""WinAction driver"""
+import subprocess
+
+import Bcfg2.Client.Tools
+from Bcfg2.Utils import safe_input
+
+
+class WinAction(Bcfg2.Client.Tools.Tool):
+    """Implement Actions"""
+    name = 'WinAction'
+    __handles__ = [('Action', None)]
+    __req__ = {'Action': ['name', 'timing', 'when', 'command', 'status']}
+
+    def RunAction(self, entry):
+        """This method handles command execution and status return."""
+        shell = True
+        shell_string = ''
+        if entry.get('shell', 'false') == 'true':
+            shell = True
+            shell_string = '(in shell) '
+
+        if not Bcfg2.Options.setup.dry_run:
+            if Bcfg2.Options.setup.interactive:
+                prompt = ('Run Action %s%s, %s: (y/N): ' %
+                          (shell_string, entry.get('name'),
+                           entry.get('command')))
+                ans = safe_input(prompt)
+                if ans not in ['y', 'Y']:
+                    return False
+            if False:
+                if entry.get('build', 'true') == 'false':
+                    self.logger.debug("Action: Deferring execution of %s due "
+                                      "to build mode" % entry.get('command'))
+                    return False
+            self.logger.debug("Running Action %s %s" %
+                              (shell_string, entry.get('name')))
+            rv = self.cmd.run(entry.get('command'), shell=shell,
+                              stdout=subprocess.PIPE,
+                              stderr=subprocess.PIPE,
+                              close_fds=False)
+            self.logger.debug("Action: %s got return code %s" %
+                              (entry.get('command'), rv.retval))
+            entry.set('rc', str(rv.retval))
+            return entry.get('status', 'check') == 'ignore' or rv.success
+        else:
+            self.logger.debug("In dryrun mode: not running action: %s" %
+                              (entry.get('name')))
+            return False
+
+    def VerifyAction(self, dummy, _):
+        """Actions always verify true."""
+        return True
+
+    def InstallAction(self, entry):
+        """Run actions as pre-checks for bundle installation."""
+        if entry.get('timing') != 'post':
+            return self.RunAction(entry)
+        return True
+
+    def BundleUpdated(self, bundle):
+        """Run postinstalls when bundles have been updated."""
+        states = dict()
+        for action in bundle.findall("Action"):
+            if action.get('timing') in ['post', 'both']:
+                if not self._install_allowed(action):
+                    continue
+                states[action] = self.RunAction(action)
+        return states
+
+    def BundleNotUpdated(self, bundle):
+        """Run Actions when bundles have not been updated."""
+        states = dict()
+        for action in bundle.findall("Action"):
+            if (action.get('timing') in ['post', 'both'] and
+                    action.get('when') != 'modified'):
+                if not self._install_allowed(action):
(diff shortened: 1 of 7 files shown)


src/lib/Bcfg2/Client/Tools/WinService.py
```diff
@@ -0,0 +1,87 @@
...
+        :returns: Bcfg2.Utils.ExecutorResult - The return value from
+                  :class:`Bcfg2.Utils.Executor.run`
+        """
+        self.logger.debug('Restarting service %s' % service.get('name'))
+        restart_target = service.get('target', 'restart')
+        return self.cmd.run(self.get_svc_command(service, restart_target), stdout=subprocess.PIPE, stderr=subprocess.PIPE, close_fds=False)
+
+    def FindExtra(self):
+        """Locate extra Windows services."""
+        return []
```
AlexanderS: Shouldn't it possible to get the list of services, that are scheduled to start automatically? Maybe something like this would work:

```
Get-WMIObject win32_service -Filter "StartMode = 'auto'" | Format-Table Name -HideTableHeaders
```

src/lib/Bcfg2/Client/__init__.py
```diff
@@ -427,10 +437,14 @@ def run(self):
...
-            if lockfile:
+            if os.path.isfile(Bcfg2.Options.setup.lockfile):
                 try:
-                    fcntl.lockf(lockfile.fileno(), fcntl.LOCK_UN)
-                    os.remove(Bcfg2.Options.setup.lockfile)
+                    if os.name == 'nt':
+                        os.unlink(Bcfg2.Options.setup.lockfile)
+                    else:
+                        lockfile = open(file, 'w')
+                        fcntl.lockf(lockfile.fileno(), fcntl.LOCK_UN)
```
AlexanderS: I think, that does not work (quote from flock(2) man page):

>  If  a process uses open(2) (or similar) to obtain more than one file descriptor for the same file, these file descriptors are treated independently by flock().  An attempt to lock the file using one of these file descriptors may be denied by a lock that the calling process has already placed via another file descriptor.

I do not know if python will cache the fd from above, but I would not rely on this.

AlexanderS:
Oh wow, thanks.

The errors reported by travis-ci, seems only style issues. That should be fixable. Would you mind to fix the obvious stuff (like line too long)?

I don't like the `os.name` name checks during the imports. Maybe we can do something like the `Compat` stuff and externalize this into an external module.

src/lib/Bcfg2/Client/Tools/WinService.py
AlexanderS: Just found the python [wmi](https://pypi.python.org/pypi/WMI/) module. Maybe we just want to rely on this, instead "parsing" the output of powershell scripts.

src/lib/Bcfg2/Client/__init__.py
```diff
@@ -201,7 +201,8 @@ def run_probe(self, probe):
         try:
             fileending = ''
             if os.name == 'nt':
-                (interpreter, fileending) = probe.attrib.get('interpreter').split('*')
+                (interpreter, fileending) =
```
AlexanderS: There is a `\` missing at the end of the line.

src/lib/Bcfg2/Client/Tools/WinService.py
chschenk: I have thought about this too, but decided against it because the WMI module is not included in the standard version of python for windows and, at least in my environment, it is extremely difficult to deploy extra pip packages onto windows servers (especially when I consider updating these packages).

src/lib/Bcfg2/Client/__init__.py
chschenk: I have implemented an completely new locking mechanism which should solve the problem on windows and on Linux. See class Flock in Utils.py

chschenk:
@AlexanderS 
What do you think of this? I have added the modules which weren't available under Windows to the Compat module and fixed the style issues

AlexanderS:
Wow, that looks great.

@solj What do you think?

src/lib/Bcfg2/Client/Tools/WinAction.py
```diff
@@ -0,0 +1,79 @@
...
+
+class WinAction(Bcfg2.Client.Tools.Tool):
+    """Implement Actions"""
+    name = 'WinAction'
+    __handles__ = [('Action', None)]
+    __req__ = {'Action': ['name', 'timing', 'when', 'command', 'status']}
+
+    def RunAction(self, entry):
+        """This method handles command execution and status return."""
+        shell = True
```
solj: Should this default to False. It seems to me like this block of code will always result in shell = True

src/lib/Bcfg2/Client/Tools/WinAction.py
```diff
@@ -0,0 +1,79 @@
...
+
+        if not Bcfg2.Options.setup.dry_run:
+            if Bcfg2.Options.setup.interactive:
+                prompt = ('Run Action %s%s, %s: (y/N): ' %
+                          (shell_string, entry.get('name'),
+                           entry.get('command')))
+                ans = safe_input(prompt)
+                if ans not in ['y', 'Y']:
+                    return False
+            if False:
```
AlexanderS: `if False:` seems wrong here.

chschenk:
You are right i have fixed the issues

odenbach:
Any chance to get this included? I would love to see this upstream.

solj:
@chschenk It looks like the exception clauses you're using in Utils.py don't work with < python 2.6. Can you modify those so they don't use the "except Exception as" clause?

PR #392, opened 2017-08-16
Use older mercurial for python2.6

AlexanderS:
> Mercurial 4.2.2 is the last release to support Python 2.6.
> Use this if you need to run Mercurial on old platforms and
> you cannot update your Python installation.

See also: https://www.mercurial-scm.org/wiki/SupportedPythonVersions


AlexanderS:
We can either use this or fix and use #297.

solj:
@AlexanderS I think #297 makes sense if it's better/newer. What are your thoughts?

AlexanderS:
I would definitely prefer #297. (And I already added this fix to it.)

AlexanderS:
Fixed by #297.

PR #395, opened 2017-11-03
travis-ci: Fix pip failure for older python versions

AlexanderS:
The older python versions only ship with an old version of pip. Since some days
SSL is required for downloading from pypi. We have to specify this explicitly
because the older pip version only try http by default.
(https://github.com/pypa/pip/issues/4817)


solj:
Merged, thanks!

PR #397, opened 2018-02-23
doc: Remove support for docutils-0.4

AlexanderS:
The compat class from sphinx for "Directive" was deprecated and removed in the
current sphinx version. So we should use the class from docutils directly.

You will need at least docutils-0.5 (from 2008) to build the documentation.


solj:
This looks good. I'm not sure why the sphinx test on 2.7 is failing.

solj:
Merged, thanks!

PR #401, opened 2018-04-27
add hierarchy plugin

xschlef:
Hi,

we use some external tools to visualize our bcfg2 data. As bcfg2 does not provide the hierarchy of the groups, this plugin adds them back. Maybe this could be part of the core, but I found that using a plugin is the cleanest variant.


solj:
Looks like the build failed due to some unused imports?

PR #402, opened 2018-04-27
WIP: Fixes some bugs with the current debian packaging

xschlef:
Hi,
the debian packages are currently broken.

Starting with debian stretch, there are no longer any south packages. south seems to be deprecated anyway so I removed the depencendy from bcfg2-web.

Also I added migrations to setup.py, as migrations were missing so the database could not be created or migrated.

This is work in progress until all bugs are fixed.


solj:
I've gone ahead and merged this. Feel free to open a new ticket with more fixes. Thanks!

PR #403, opened 2018-05-18
POSIX: Don't follow symlinks when changing owner

weiss:
Don't let the client follow symbolic links when changing the owner of a path.


solj:
Merged, thanks!

PR #404, opened 2018-06-27
Set default SSL protocol version to xmlrpc/ssl

xschlef:
This is the secure choice with modern distributions. xmlrpc/tls1 limits the communication to TLSv1, while xmlrpc/ssl uses ssl.PROTOCOL_SSLv23. This protocol flag indicates that SSLv2 and SSLv3 could be used, but they are usually disabled in current versions of OpenSSL and allow clients to use the highest supported TLS version.

The best way would be to disable SSLv2 / SSLv3 contexts for the socket entirely (which is done by default in python 3)

See:
https://docs.python.org/2/library/ssl.html#socket-creation

The protocol name ssl.PROTOCOL_SSLv23 is deprecated starting with python 3.6 and is renamed to ssl.PROTOCOL_TLS.

### DIFF
--- a/src/lib/Bcfg2/Server/SSLServer.py
+++ b/src/lib/Bcfg2/Server/SSLServer.py
@@ -335,7 +335,7 @@ class XMLRPCServer(SocketServer.ThreadingMixIn, SSLServer,
     """ Component XMLRPCServer. """
 
     def __init__(self, listen_all, server_address, RequestHandlerClass=None,
-                 keyfile=None, certfile=None, ca=None, protocol='xmlrpc/tlsv1',
+                 keyfile=None, certfile=None, ca=None, protocol='xmlrpc/ssl',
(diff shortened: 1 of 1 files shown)


solj:
@xschlef I think the main issue here is that for users on python 2.6 (which we still support), this will cause their connection to be insecure by default.

https://docs.python.org/2.6/library/ssl.html#ssl.PROTOCOL_SSLv23

Perhaps we could have this be set to PROTOCOL_TLS for newer python and fallback to PROTOCOL_TLSv1 for older clients? According to the docs, it looks like SSLv23 is deprecated anyway.

xschlef:
@solj users of such legacy systems that are using python 2.6 + openssl in an old enough version to still support SSLv2/3 should be aware that they have potential security problems in their infrastructure.

PR #405, opened 2018-06-27
Additional fixes for debian packaging

xschlef:
This merges most of the fixes from debian packaging as well as creating debian specific unit files for systemd.


xschlef:
debian is handling the packaging

PR #406, opened 2018-10-12
Fix server shutdown

AlexanderS:
Fix shutdown of the server.


solj:
:+1:  Looks good to me! Thanks!

PR #408, opened 2019-05-07
Lint: Import lint plugins with package name

AlexanderS:
When importing packages we should not pollute the global modules
namespace and always import new modules with their package name.

Without this fix, I could not import the module `Crypto.Util` in a TemplateHelper while linting, because `Bcfg2.Server.Lint.Crypto` was already imported as `Crypto`.


solj:
:+1: Agreed. Merged, thanks!

PR #409, opened 2019-05-07
Fix testsuite

AlexanderS:
This will fix the testsuite:

* Python 2.6 requires an older version of lxml and PyYAML
* python-mock documentation was removed from voidspace.org.uk
* boto requires google_compute_engine (at least in the python2.6 environment of travis-ci)


solj:
:+1: Merged, thanks!

PR #411, opened 2019-06-17
Reporting: Fix add_url_filter fallback

AlexanderS:
django-1.8 deprecated referencing views as strings in the url(), so we
need to import the fallback view and referencing the view function
directly.


solj:
Merged, thanks!

PR #412, opened 2019-07-18
Do not encode py3 strings to utf-8 again.

xschlef:
Do not encode py3 strings to utf-8 again. Compat.unicode matches strings and unicode in py3.


solj:
Merged, thanks!

PR #418, opened 2019-11-26
Add client support for python 3.5+

### LINKED ISSUE #415: Python 3.7 support
Using debian buster and python3.7, I cannot connect to the bcfg2 server as python3.7 deprecated the following behavior:
`Unknown failure: SSLSocket does not have a public constructor. Instances are returned by SSLContext.wrap_socket(). (TypeError)
Fatal error: Failed to declare version: Unknown failure: SSLSocket does not have a public constructor. Instances are returned by SSLCont
ext.wrap_socket(). (TypeError)`

See also:
https://github.com/python/cpython/commit/89c2051a554d2053ac87b0adbf11ed0f1bb65db3

xschlef:
Starting with python 3.7 ssl.SSLSocket was removed. Simply calling wrap_socket is the proposed fix here.
Starting with python 3.5 ssl.PROTOCOL_TLSv1 was removed, so we switch to PROTOCOL_TLS, which should result in the highest possible TLS connection. Modern OSes do no longer support SSLv2/3 so it is okayish to ignore those.

This should fix #415

### DIFF
--- a/src/lib/Bcfg2/Client/Proxy.py
+++ b/src/lib/Bcfg2/Client/Proxy.py
@@ -198,6 +198,9 @@ class SSLHTTPConnection(httplib.HTTPConnection):
             ssl_protocol_ver = ssl.PROTOCOL_SSLv23
         elif self.protocol == 'xmlrpc/tlsv1':
             ssl_protocol_ver = ssl.PROTOCOL_TLSv1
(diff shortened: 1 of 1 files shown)


solj:
This looks good to me although we may want to consider moving this logic into Compat.py in order to fix the travis failures.

PR #419, opened 2019-11-26
Fix default timezone in debian buster

xschlef:
bcfg2-web will fail without setting a valid time zone in newer django versions. This fix initializes the time zone as utc instead of None, to fix most setups without directly breaking functionality. This can be fixed without changing the code by adding this to the config file.
```
# /etc/bcfg2-web.conf
[reporting]
TIME_ZONE = <valid_tz>
```

using UTC is recommended by django: https://riptutorial.com/django/example/3107/setting-the-timezone


solj:
Merged, thanks!

PR #420, opened 2021-01-22
Update Travis-CI configuration

AlexanderS:
Travis-CI had deprecated the old container infrastructure and allows to
use sudo again. This also bumps the ubuntu version so something more
current.

We are trying to use packages for most dependencies, when using the
system python version (2.7 for bionic).


solj:
@AlexanderS do you think it's still worth the effort to support python 2.6? Given that upstream has dropped support, it may make life easier to move forward with recent versions.

